PR #25, opened 2026-04-09
fix: click focus on panes running TUIs with mouse reporting

othavi0:
## Summary

Two related bugs in `AppState::handle_mouse` that broke mouse clicks on panes running a TUI with mouse reporting enabled. Claude Code reproduces both. Codex and plain shells do not, because they don't toggle any of the DECSET mouse modes (1000/1002/1003), so the code path that caused the bug never fires for them.

On a Claude Code pane:

- Left-click did nothing to focus. The pane border stayed on whatever was previously focused.
- Right-click (after the left-click fix) focused the pane but never opened the herdr context menu. Split, fullscreen, close were unreachable by mouse.

## Root cause

Both handlers called `forward_pane_mouse_button` first and early-returned when the forward succeeded. Forwarding succeeds whenever `mouse_protocol_mode != None`, i.e. whenever the inner TUI has enabled any DECSET mouse mode. So the rest of each handler, including the focus update and the context menu creation, never ran for Claude Code panes.

`handle_terminal_wheel` already did it the right way: focus first, forward second. That's why scrolling "accidentally" updated focus as a side-effect, and why the bug looked click-specific.

## The fix

Two commits, one per bug.

1. `fix: focus pane on left-click even when mouse reporting is active`
   Reorder `Down(MouseButton::Left)` so `focus_pane` and the mode transition run before the forward, matching the wheel handler. The forward still runs afterwards when the inner TUI wants the click.

2. `fix: always open pane context menu on right-click`
   Stop forwarding `Down(MouseButton::Right)` to the pane runtime entirely. Right-click on a pane belongs to herdr, not to the inner content, same as Ghostty / Alacritty / Kitty / iTerm2. Also dropped `Up(MouseButton::Right)` and `Drag(MouseButton::Right)` from the middle/right forward branch so the inner TUI never sees an up or drag without a matching down. Middle button stays because some protocols still use it for paste.

## Tests

One regression test per commit. Each builds a two-pane workspace and installs a `PaneRuntime` on the target pane with mouse reporting enabled. The bytes `\x1b[?1002h` are written to the runtime at construction, which goes through Ghostty's real VT parser, so `mouse_protocol_mode` ends up as an actual `ButtonMotion` rather than a mock. The test fires a `Down(Left)` or `Down(Right)` and asserts focus moved. The right-click test also asserts `Mode::ContextMenu` was entered and `context_menu` is populated.

I did the negative proof on both: revert the fix, run the test, confirm it fails, restore. Worth mentioning: the first version of the right-click test only checked focus and would have passed against broken code. That's how the context-menu half of the bug slipped past the original attempt in this PR. The tightened assertions catch it now.

`cargo build`, `cargo fmt --check`, `cargo test` all clean (453 unit + 22 integration).

## Manual verification

1. Run Claude Code in a pane.
2. `Ctrl+B` → `V` to split vertically.
3. Left-click on the Claude Code pane → focus moves correctly.
4. Right-click on the Claude Code pane → context menu opens with split/fullscreen/close entries.
5. Invoke a context menu action (e.g. split vertical) → it acts on the Claude Code pane, not on the previously-focused shell pane.
6. Repeated the same with Codex and a plain shell as controls → no behaviour change (they were already working).
7. Scroll wheel, drag-to-select text, scrollbar clicks, tab/sidebar clicks, and split-border drag resize all verified to still work.

### DIFF
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -2336,10 +2336,7 @@ impl AppState {
                 }
             }
 
-            MouseEventKind::Up(MouseButton::Middle)
-            | MouseEventKind::Up(MouseButton::Right)
-            | MouseEventKind::Drag(MouseButton::Middle)
-            | MouseEventKind::Drag(MouseButton::Right)
+            MouseEventKind::Up(MouseButton::Middle) | MouseEventKind::Drag(MouseButton::Middle)
                 if !in_sidebar =>
             {
                 if let Some(info) = self.pane_mouse_target(mouse.column, mouse.row).cloned() {
@@ -2440,9 +2437,7 @@ impl AppState {
 
             MouseEventKind::Down(MouseButton::Right) if !in_sidebar => {
                 if let Some(info) = self.pane_mouse_target(mouse.column, mouse.row).cloned() {
-                    if self.forward_pane_mouse_button(&info, mouse) {
-                        return None;
-                    }
+                    self.focus_pane(info.id);
                     self.context_menu = Some(ContextMenuState {
                         kind: ContextMenuKind::Pane,
                         x: mouse.column,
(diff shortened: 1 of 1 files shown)


Bas-Martin:
Implemented in branch custom: persistent SSH targets, remote task/project metadata, isolated remote worktree provisioning, retry, lifecycle routing, and teardown handling.

int19h:
Design update (2026-09-12): the focused finding-7 Grok manifest amendment revision 2 has received unconditional Opus-xhigh design approval. The approved arrangement is a staged pair: distribution `2026.07.16.1.1` / engine 2 with the neutral completion marker, and bundled `2026.09.12.1` / engine 3 with the same marker plus `background_work_chip_working`. The version ordering keeps pre-#25 engine-3 clients on their newer bundled manifest while preserving the engine-2 update channel. The existing Opus-xhigh implementer has been authorized to incorporate this amendment with corrections for review findings 1-6 and produce a successor candidate. No merge, push, publication, installation, or deployment is authorized.

int19h:
Exact-head review update: the original designer reviewed intermediate head `ff561fd2` and requested changes. Corrections for prior findings 1, 2, 3, 5, and 6 are substantively accepted. Remaining blockers are: require a race-free pre-submit compaction cursor rather than an optional entry baseline; probe after entry through the frozen canonical pane/terminal identity so disappearance or replacement yields normal `unknown(target_identity_changed)`; and implement the approved finding-7 staged Grok manifest amendment. The same Opus-xhigh implementer is correcting these. No merge, push, publication, installation, or deployment is authorized.

int19h:
Candidate status: original-designer exact-head review APPROVED `87c72a11c9a21eb15b2a7fe9580226a37cef67ef` with no blocking findings. Reported checks: fmt/clippy clean; 2959 tests pass, with one known environment-dependent failure reproduced identically at base; manifest gate and focused maintenance/architecture suites pass. The candidate remains parked atop issue #29 base `1252baee...`; it has not been merged, pushed, published, installed, or deployed. Non-blocking cleanup note: one application-level fixture retains a PTY receiver but its claimed empty-channel assertion is actually evidence-provenance checking; the deterministic waiter seam itself is read-only and this does not block approval.

int19h:
Final re-anchor: the original designer approved test-only successor `6cb5e2d06f0c4eda6d5fadeeb55384299ae01968` and recommends keeping it as a separate commit atop `87c72a11`. The delta replaces an unrelated evidence-provenance assertion with a direct empty-PTY-channel assertion; production behavior is unchanged. This is now the authoritative approved and parked issue #25 head. Nothing has been merged, pushed, published, installed, or deployed.

int19h:
Additional pre-install field correlation from Smusni I94 v5: three ack-required completion messages were found later at status without visible prompts. Sender-side dispositions concretely record Kimi and Astra nudges as skipped because the PM was classified `working`; the Grok sender receipt was not preserved in the inspected message/ack records and remains unclassified. A nearby Fable completion did notify while the PM was idle. This is consistent with the state-dependent suppression addressed by the approved candidate, not a post-fix regression: approved head `6cb5e2d0` remains uninstalled.

PR #27, opened 2026-04-09
fix: extend selection when wheel-scrolling during a drag

othavioquiliao:
Bug fix I hit copying shell output. `just ci` clean. Flagging one opinionated call near the end.

## what's broken

Hold left, drag to select text in a pane, then roll the wheel to keep pulling the selection past the viewport. The selection shrinks instead of growing. Shells, Codex, anything that leaves mouse reporting off. Claude Code panes don't reproduce because the click goes to the TUI before a selection ever exists.

`scroll_selection_with_wheel` scrolled the pane and then called `update_selection_cursor(mouse.column, mouse.row)`. The mouse sits wherever the user stopped dragging, usually the viewport edge, so `Selection::drag` recomputes the absolute row as `new_viewport_top + clamped_viewport_row`. As `new_viewport_top` moves with the scroll, the cursor chases the scroll instead of extending the selection.

## the fix

Two commits plus a cleanup:

1. New `Selection::extend_in_scroll_direction(scroll_up, rows)` in `src/selection.rs`. `pub(crate)` helper that moves the top or bottom endpoint by `rows`, picking the endpoint with the same row+column ordering as `Selection::ordered`. The column tie-break matters for single-row right-to-left drags.

2. `scroll_selection_with_wheel` stops feeding mouse coordinates into the selection. Measures `offset_from_bottom` before and after the scroll, passes the real delta to the new helper. If the pane hit the scrollback boundary and couldn't scroll, the selection doesn't extend either.

3. `chore:` commit trimming narrative comments I left behind during iteration.

## the opinionated half

The stored `anchor` can now be mutated by the wheel. When `anchor` is the top endpoint and the user scrolls up, `anchor.0` decreases; when it's the bottom and they scroll down, `anchor.0` increases. The "anchor is where you first pressed" invariant no longer holds.

I chose this because it's what Alacritty does (their issue #1600 fixed the same bug with the same trade-off), and the alternative is a separate `wheel_offset` field threaded through `ordered`, `contains`, `drag`, and every caller. Too much state for a rarely-hit edge case.

If you'd rather keep `anchor` immutable and push the extra state, say so and I'll rework it. Mouse ergonomics is opinionated so this is your call, not mine.

## does this change UI / interaction / workflow?

Interaction: yes, but only for a case that was already broken. Wheel during an in-progress selection now extends instead of shrinks. Matches what I see in Alacritty, Ghostty, Kitty, iTerm2.

Everything else stays: keybinds, layout, sidebar, scrollbar, drag select outside the wheel path, auto-scroll when the drag cursor leaves the viewport, copy on release, OSC 52 write.

## tests

7 unit tests on the helper covering both scroll directions across both endpoint owners, the single-row right-to-left tie-break, the `Anchored → Dragging` phase transition, and zero-saturation. 2 end-to-end regression tests in `input.rs` for forward drag + ScrollUp and reverse drag + ScrollDown. Each fails on the old code and passes on the new one. I reverted each fix locally with the tests in place to confirm before pushing. `just ci` clean, 462 unit + 22 integration tests.

## honest caveat

If a `Drag` event arrives after the wheel extension (user moves the mouse before releasing), the existing drag handler recomputes the cursor from the physical position and can pull the wheel-extended endpoint back. Clean flow is unaffected. Leaving it as a follow-up.

### DIFF
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -3051,10 +3051,8 @@ impl AppState {
 
         self.focus_pane(pane_id);
 
-        // Measure the actual scroll delta: wheel requests LINES_PER_NOTCH,
-        // but the pane clamps at top/bottom of scrollback. Passing the
-        // requested notch straight into the selection math would over-extend
-        // the selection at those boundaries.
+        // Use the actual scroll delta — the pane clamps at the scrollback
+        // edges, so the notch we requested may not match what scrolled.
         let before_offset = self
             .pane_scroll_metrics(pane_id)
             .map(|m| m.offset_from_bottom);
@@ -4508,11 +4506,6 @@ mod tests {
 
     #[tokio::test]
     async fn wheel_scroll_during_drag_extends_selection_beyond_viewport() {
-        // Regression: dragging a selection to the bottom of the viewport and
-        // then scrolling the wheel upward used to shrink the selection because
-        // the wheel handler recalculated the cursor from the stuck mouse
-        // position combined with the new scroll metrics. The fix extends the
-        // selection by the effective scroll delta instead.
         let mut app = app_for_mouse_test();
         let mut ws = Workspace::test_new("test");
         let pane_id = ws.tabs[0].root_pane;
@@ -4542,8 +4535,6 @@ mod tests {
         let bottom_row = info.inner_rect.y + info.inner_rect.height - 1;
         let col = info.inner_rect.x + 2;
 
-        // Click at the top of the viewport and drag to the bottom. This
-        // selects the full viewport height.
         app.handle_mouse(mouse(MouseEventKind::Down(MouseButton::Left), col, top_row));
         app.handle_mouse(mouse(
             MouseEventKind::Drag(MouseButton::Left),
@@ -4560,8 +4551,6 @@ mod tests {
             "sanity: drag should select multiple rows (got {before_rows})"
         );
 
-        // Scroll wheel up while still holding — should extend the selection
-        // upward into the scrollback, not shrink it.
         app.handle_mouse(mouse(MouseEventKind::ScrollUp, col, bottom_row));
 
         let selection_after = app.state.selection.as_ref().expect("selection after wheel");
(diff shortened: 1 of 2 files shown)


ogulcancelik:
i'll release a new version, can you see if this issue still exist? because i can't reproduce this

othavioquiliao:
I made the update here and ran a few tests, looks like the issue is resolved.

It's a bit odd though. Yesterday, selecting text and then scrolling would cause the selection to disappear. None of the changes in PR #25 or #30 should have affected this behavior.

For context, I worked on #30 in my fork, which already included the fix from this PR. Still, that shouldn't have impacted it in this way.

Anyway, it seems to be fixed in v0.4.7.

othavioquiliao:
Going to dig into this more when I get a chance, probably later today. I'll come back with something concrete once I do.

othavioquiliao:
Quick follow-up with something concrete. I was wrong earlier about v0.4.7 fixing anything on the selection path. The "disappears on scroll" symptom I first reported was already fixed by your own 953ff58 in v0.4.6. I was still on v0.4.5 yesterday when I hit it, so updating to v0.4.7 made the symptom go away because v0.4.6 already shipped the fix.

I ran the two regression tests in this PR (`wheel_scroll_during_drag_extends_selection_beyond_viewport` and `wheel_scroll_down_during_reverse_drag_extends_selection`) against a clean checkout of `upstream/master` at `aa6da3c`, without the fix applied. Both fail with the exact behavior they're testing against:

```
---- wheel_scroll_during_drag_extends_selection_beyond_viewport ----
top row should move upward after scroll up: before=47 after=47

---- wheel_scroll_down_during_reverse_drag_extends_selection ----
top row should stay anchored: before=37 after=40
```

So the code-level bug is real. What's subtle is the visual symptom: the wheel handler uses `LINES_PER_NOTCH = 3`, so each notch shifts the selection by three rows in the wrong direction. Small enough that most gestures read as "scroll with selection works", which is probably why you couldn't reproduce manually. Your own test (`wheel_scroll_keeps_in_progress_selection_and_extends_it`) covers the symmetric case where the mouse stays at the top during a scroll up, which the buggy code happens to get right by coincidence.

Also just force-pushed the branch: rebased on top of current `upstream/master`, and added a small refactor commit that shares the test setup across the three wheel-scroll tests (the two new ones and your existing one). All regression tests still pass. Your call on how to handle the PR from here.

PR #30, opened 2026-04-10
fix: forward OSC 52 clipboard writes from pane children (#29)

### LINKED ISSUE #29: copying text does not work in ampcode
# Summary

Copying text doesn't work when I'm using Amp.
Image

I'd see that Amp's toast about the clipboard, but nothing gets copied to my Clipboard. This doesn't happen on Opencode, Claude Code, etc. 

# Different ways to reproduce
- Highlight text and wait for Amp's toast
- Highlight text and press cmd + c etc

othavioquiliao:
## tl;dr

Fixes #29. Select text inside Amp in a herdr pane, wait for the "copied" toast, paste somewhere, get nothing. Any program that writes OSC 52 from inside a pane has the same problem. Amp just happens to be the first agent on the list to hit it.

## What's happening

herdr embeds `libghostty-vt` as a full VT emulator per pane. Not a passthrough. Every escape sequence the child writes lands in the emulator and stops there unless herdr explicitly re-emits it.

The readonly stream handler at `vendor/libghostty-vt/src/terminal/stream_terminal.zig:259-266` drops several OSC actions it doesn't implement, and OSC 52 is one of them:

```zig
// Have no terminal-modifying effect
.report_pwd,
.show_desktop_notification,
.progress_report,
.clipboard_contents,   // <-- this one
.title_push,
.title_pop,
=> {},
```

So when Amp writes `\x1b]52;c;<base64>\x1b\\` into its PTY, the bytes reach `process_pty_bytes()`, the emulator parses the sequence, recognizes it, and discards it. Nothing reaches herdr's stdout, so the host terminal never sees it. Amp's toast is honest from its side: it wrote the bytes. They just died on the way up.

herdr's own selection already dodges this by writing OSC 52 straight to `std::io::stdout()` in `src/selection.rs:201-208`, skipping the emulator. That works while the user is dragging with herdr's own handler. But when a program enables mouse reporting (DECSET 1002), herdr forwards mouse events to the child, the child does its own selection, and then the copy path runs into the wall. Claude Code, opencode, and Codex don't trigger this because they don't turn mouse reporting on for selection. Amp does. The bug is structural; Amp just got there first.

## What the fix does

Added `Osc52Forwarder` in `src/pane.rs`. A small state machine (`Ground` / `Escape` / `OscBody` / `OscEscape`) watches raw PTY bytes in `process_pty_bytes` and reconstructs complete `ESC ] 52 ; c ; <base64> (BEL|ST)` sequences. It handles chunks split across `read()` calls, since PTY reads don't respect sequence boundaries. It only forwards writes to the standard clipboard: `52;c;` and `52;;` (ghostty treats the empty selector as `c`, see `clipboard_operation.zig:24-46`). Queries (`?`) and other selectors get dropped on purpose, since herdr has no path to reply to a query and nothing in the wild targets the rest.

Recognized sequences ride back on `ProcessBytesResult.clipboard_writes`. The reader task drains them and fires a new `AppEvent::ClipboardWrite`, and `handle_internal_event` writes the bytes to stdout.

One thing worth flagging, because I got it wrong on my first pass and the plan-review caught it: **don't write to stdout from the PTY reader task.** `process_pty_bytes` runs on `spawn_blocking`. The main loop is the thread that owns stdout, so `terminal.draw`, `query_host_terminal_theme`, `selection::write_osc52`, and the tmux keyboard-mode toggles all write from there. A direct stdout write from the reader would race ratatui's render output and cause intermittent terminal corruption at exactly the seam this PR is trying to close. Routing through `AppEvent::ClipboardWrite` keeps everything serialized on the main loop, same as every other stdout write in the codebase.

A couple of smaller decisions:

Payload cap at 256 KiB. Base64 at that size is about 192 KiB of raw text, which covers real source-file copies without leaving the buffer unbounded when the stream contains garbage. The upstream stream parser has a comment noting OSC 52 can be "arbitrarily large" (`stream.zig:437-446`), so I erred generous.

`try_send` on the events channel, not `block_on(send)`. If the main loop is saturated, dropping a clipboard write is the better failure mode than stalling the PTY reader. A lost copy the user can redo. A stalled reader freezes the whole pane.

Vendored `libghostty-vt` stays untouched. I looked at adding a `clipboard_write` callback to its C API in `vendor/.../c/terminal.zig:41-80` and decided it wasn't worth it: modify vendored Zig, add new FFI, rebuild the emulator, all to replicate what a small Rust observer already does from outside. If herdr ever needs OSC 52 read (host responding to a child query), that calculus changes, but we're not there yet.

## Files

- `src/events.rs` — new `AppEvent::ClipboardWrite { bytes: Vec<u8> }` variant
- `src/pane.rs` — `Osc52Forwarder`, wire-up in `GhosttyPaneCore` and `process_pty_bytes`, `try_send` in the reader task, 15 unit tests
- `src/app/mod.rs` — `handle_internal_event` early-returns on `ClipboardWrite` after writing to stdout
- `src/app/actions.rs` — no-op arm in `handle_app_event` so the match stays exhaustive

## Testing

`cargo test` runs 477 unit + 11 API + 11 CLI wrapper tests, all green. The 15 new forwarder tests cover BEL and ST terminators, empty-selector form, query rejection, non-`c` selector rejection, split sequences at three different cut points (mid-payload, just before BEL, between ESC and `\` of the ST terminator), payload-size limit with recovery, garbage-in recovery, ESC-not-followed-by-backslash inside the body, multiple sequences in one chunk, and drain-clears-pending.

`cargo clippy` is clean on the touched files. Pre-existing warnings in `build.rs` and `workspace.rs` are unrelated.

Manual:

- Ran `amp` in a herdr pane, dragged to select text, waited for the toast, pasted into another app. Works.
- Regression sanity: plain `bash` pane with no mouse reporting. herdr's native selection path still copies via `selection::write_osc52` as before.

Thanks for building herdr, it's become my daily driver for juggling agents. Happy to iterate on anything in this PR.

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -453,6 +453,9 @@ impl AppState {
                 .update_pane_state(pane_id, |pane| pane.release_agent(&source, agent))
                 .into_iter()
                 .collect(),
+            // Intercepted in App::handle_internal_event before reaching this
+            // dispatch; never touches AppState.
+            AppEvent::ClipboardWrite { .. } => Vec::new(),
         }
     }
 
--- a/src/app/input.rs
+++ b/src/app/input.rs
@@ -2887,13 +2887,43 @@ impl AppState {
             return false;
         }
         let pane_id = selection.pane_id;
-        self.focus_pane(pane_id);
-        match mouse.kind {
-            MouseEventKind::ScrollUp => self.scroll_pane_up(pane_id, LINES_PER_NOTCH),
-            MouseEventKind::ScrollDown => self.scroll_pane_down(pane_id, LINES_PER_NOTCH),
+
+        let scroll_up = match mouse.kind {
+            MouseEventKind::ScrollUp => true,
+            MouseEventKind::ScrollDown => false,
             _ => return false,
+        };
+
+        self.focus_pane(pane_id);
+
+        // Measure the actual scroll delta: wheel requests LINES_PER_NOTCH,
+        // but the pane clamps at top/bottom of scrollback. Passing the
+        // requested notch straight into the selection math would over-extend
+        // the selection at those boundaries.
+        let before_offset = self
+            .pane_scroll_metrics(pane_id)
+            .map(|m| m.offset_from_bottom);
+        if scroll_up {
+            self.scroll_pane_up(pane_id, LINES_PER_NOTCH);
+        } else {
+            self.scroll_pane_down(pane_id, LINES_PER_NOTCH);
         }
-        self.update_selection_cursor(pane_id, mouse.column, mouse.row);
+        let after_offset = self
+            .pane_scroll_metrics(pane_id)
+            .map(|m| m.offset_from_bottom);
+
+        let effective_delta = match (before_offset, after_offset) {
+            (Some(before), Some(after)) if scroll_up => after.saturating_sub(before) as u32,
+            (Some(before), Some(after)) => before.saturating_sub(after) as u32,
+            _ => 0,
+        };
+
+        if effective_delta > 0 {
+            if let Some(selection) = self.selection.as_mut() {
+                selection.extend_in_scroll_direction(scroll_up, effective_delta);
+            }
+        }
+
         true
     }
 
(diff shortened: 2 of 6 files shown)


ogulcancelik:
fixed with some additions on top of your commits ty

PR #45, opened 2026-04-21
fix: resolve transparent bg before inverse swap to prevent invisible text

### LINKED ISSUE #44: Inverse/reverse video text is invisible when background is transparent
## Bug description

When pasting text in herdr, all colors appear white (both text and highlight), making the content unreadable. This happens because shells like zsh and fish use **inverse/reverse video** (SGR 7) to highlight pasted text during bracketed paste mode.

## Root cause

In `src/pane/terminal.rs`, the render pipeline intentionally sets the default background to `None` (transparent) when the ghostty terminal's background matches the host terminal's background. This is correct for normal rendering — it lets the host terminal's real background show through.

However, when a cell has the `inverse` attribute set, `ghostty_cell_style` swaps `fg` and `bg`:

```rust
if style_data.inverse {
    std::mem::swap(&mut fg, &mut bg);
}
```

When `bg` is `None` (transparent), after the swap:
- `fg` becomes `None` → ratatui renders as `Color::Reset` → host terminal uses its **default foreground** color
- `bg` becomes the original fg color (e.g. `#cdd6f4`)

Since the host terminal's default foreground is the same as the color now in `bg`, the text is invisible (same color on same color).

## Steps to reproduce

1. Run herdr in a terminal with a dark theme (e.g. Catppuccin Mocha)
2. Open a pane with zsh or fish shell
3. Paste any text — the pasted text appears as an unreadable block where text and background are the same color

## Expected behavior

Inverse text should swap fg and bg correctly: dark text on a light background (or vice versa), producing readable highlighted text.

## Fix

Before swapping for inverse, resolve the transparent `None` background to the actual terminal background color so the swap produces correct contrasting colors.

EYH0602:
## Summary

- Fixes inverse/reverse video rendering when the default background is transparent (matches host terminal), which caused pasted text to be invisible in shells that use SGR 7 for bracketed paste highlighting
- Adds `resolved_bg` parameter to `ghostty_cell_style` that carries the actual terminal background color for use during inverse swap
- Adds test verifying inverse text produces correct contrasting fg/bg colors

Fixes #44

## Details

The render pipeline sets `default_bg = None` when the ghostty terminal's background matches the host terminal's background, so ratatui leaves it transparent. When a cell has the `inverse` attribute (SGR 7), the code swaps `fg` and `bg`. But swapping with `None` produces `fg = None` (Color::Reset), which the host terminal renders as its default foreground — the same color as the new `bg`, making text invisible.

The fix resolves `None` to the actual background color before the swap, so inverse correctly produces dark-on-light (or light-on-dark) text.

## Test plan

- [ ] `cargo test --bin herdr pane::terminal` — new test `render_inverse_text_swaps_fg_and_resolved_bg_when_bg_is_transparent` verifies the fix
- [ ] Manual: paste text in a zsh/fish pane inside herdr — pasted text should be readable with inverse highlighting

### DIFF
--- a/src/pane/terminal.rs
+++ b/src/pane/terminal.rs
@@ -554,6 +554,7 @@ impl GhosttyPaneTerminal {
         let colors = render_state.colors().ok();
         let default_bg = colors.and_then(|c| ghostty_default_bg(c.background, host_theme));
         let default_fg = colors.map(|c| ghostty_color(c.foreground));
+        let resolved_bg = colors.map(|c| ghostty_color(c.background));
 
         let mut row_iterator = match crate::ghostty::RowIterator::new() {
             Ok(iterator) => iterator,
@@ -580,7 +581,7 @@ impl GhosttyPaneTerminal {
                 let mut x = 0u16;
                 while x < area.width && cells.next() {
                     let wide = cells.wide().unwrap_or(crate::ghostty::CellWide::Narrow);
-                    let style = ghostty_cell_style(&cells, default_fg, default_bg);
+                    let style = ghostty_cell_style(&cells, default_fg, default_bg, resolved_bg);
                     let symbol = match ghostty_buffer_symbol_into(
                         &cells,
                         wide,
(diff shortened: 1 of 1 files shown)


ogulcancelik:
looks like tests are broken for ci env, will fix and merge soon

Issue #53, opened 2026-04-25
Support named persistent sessions


fbettag:
## Current behavior

Herdr currently has one persistent session per user/config directory.

Running `herdr` starts or attaches to the default background server. That works well for a single long-lived workspace set, but it means unrelated contexts share the same server, socket, restored session state, and workspace list.

For example, if I want one Herdr session for work and separate sessions for side projects, I currently need to manually isolate config/socket paths with environment variables, which is not discoverable or ergonomic.

## Proposed behavior

Add an optional named session flag:

```bash
herdr --session work
herdr --session side-projects
herdr --session work server stop
herdr --session side-projects workspace list
```

Omitting `--session` would keep today's default behavior exactly as-is.

Named sessions would use separate runtime/session state, while continuing to share the normal Herdr config file:

```text
~/.config/herdr/session.json
~/.config/herdr/sessions/work/session.json
~/.config/herdr/sessions/work/herdr.sock
~/.config/herdr/sessions/work/herdr-client.sock
~/.config/herdr/sessions/side-projects/session.json
~/.config/herdr/sessions/side-projects/herdr.sock
~/.config/herdr/sessions/side-projects/herdr-client.sock
```

So:

- `herdr` attaches to the default session
- `herdr --session work` attaches to or starts the `work` session
- `herdr --session side-projects` attaches to or starts a separate session for side projects
- CLI/API wrapper commands can target a named session with the same flag

Session names can be limited to a conservative filesystem-safe set, for example ASCII letters, numbers, `.`, `_`, and `-`.

## Why this belongs in Herdr

Herdr is already built around persistent agent workspaces. Named sessions would extend that persistence model without changing the core UI or interaction pattern.

The main use case is keeping unrelated agent contexts separated:

- work vs side projects
- client A vs client B
- long-running background agents vs an experimental scratch session
- a stable default session plus temporary focused sessions
- multiple ongoing personal projects grouped under a dedicated `side-projects` session

This fits Herdr's existing model better than requiring users to manually override socket paths or config homes.

## UI / workflow impact

This can be CLI-only and does not require new in-app UI.

The default workflow remains unchanged:

```bash
herdr
```

The new workflow is opt-in:

```bash
herdr --session work
herdr --session side-projects
```

No new visual language, modal, sidebar behavior, or interaction pattern is required.

## Prototype

I implemented a working prototype here:

https://github.com/fbettag/herdr/tree/issue/named-sessions

The prototype:

[...]

PR #57, opened 2026-04-27
Add named persistent sessions

fbettag:
## Summary

- add named persistent session namespaces through global `--session <name>` and `HERDR_SESSION`
- add `herdr session list`, `herdr session stop <name>`, and `herdr session delete <name>`
- preserve the default session paths and keep `HERDR_SOCKET_PATH` as a low-level exact override unless CLI `--session` is explicit
- document sessions as runtime/socket namespaces rather than workspace replacements

Refs #53.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `python3 -m unittest scripts.test_changelog scripts.test_vendor_libghostty_vt`
- `PATH=/tmp/zig-0.15.2-herdr:$PATH cargo +stable test session::tests -- --test-threads=1`
- `PATH=/tmp/zig-0.15.2-herdr:$PATH cargo +stable test socket_path -- --test-threads=1`
- `PATH=/tmp/zig-0.15.2-herdr:$PATH cargo +stable test client_socket_path -- --test-threads=1`
- `PATH=/tmp/zig-0.15.2-herdr:$PATH cargo +stable test -- --test-threads=1`
- `PATH=/tmp/zig-0.15.2-herdr:$PATH cargo +stable build --release`

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -1,0 +1,136 @@
+use std::path::PathBuf;
+
+pub const SESSION_ENV_VAR: &str = "HERDR_SESSION";
+
+const MAX_SESSION_NAME_LEN: usize = 64;
+
+pub fn configure_from_args(args: &[String]) -> Result<Vec<String>, String> {
+    let mut cleaned = Vec::with_capacity(args.len());
+    if let Some(program) = args.first() {
+        cleaned.push(program.clone());
+    }
+
+    let mut requested_session = None;
+    let mut index = 1;
+    while index < args.len() {
+        let arg = &args[index];
+        if arg == "--session" {
+            let Some(value) = args.get(index + 1) else {
+                return Err("missing value for --session".to_string());
+            };
+            requested_session = Some(value.clone());
+            index += 2;
+            continue;
+        }
+        if let Some(value) = arg.strip_prefix("--session=") {
+            requested_session = Some(value.to_string());
+            index += 1;
+            continue;
+        }
+
+        cleaned.push(arg.clone());
+        index += 1;
+    }
+
+    if let Some(session) = requested_session {
+        validate_name(&session)?;
+        std::env::set_var(SESSION_ENV_VAR, session);
+    } else if let Ok(session) = std::env::var(SESSION_ENV_VAR) {
+        validate_name(&session)?;
+    }
+
+    Ok(cleaned)
+}
+
+pub fn active_name() -> Option<String> {
+    std::env::var(SESSION_ENV_VAR)
+        .ok()
+        .filter(|name| validate_name(name).is_ok())
+}
+
+pub fn data_dir() -> PathBuf {
+    let config_dir = crate::config::config_dir();
+    match active_name() {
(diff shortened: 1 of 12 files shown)


src/session.rs
```diff
@@ -0,0 +1,483 @@
...
+
+    if let Some(session) = requested_session {
+        let session = normalize_name(&session)?;
+        if let Some(session) = session {
+            std::env::set_var(SESSION_ENV_VAR, session);
+        } else {
+            std::env::remove_var(SESSION_ENV_VAR);
+        }
+        EXPLICIT_SESSION_REQUESTED.store(true, Ordering::Relaxed);
+    } else if let Ok(session) = std::env::var(SESSION_ENV_VAR) {
```
ogulcancelik, line 63: this validates `HERDR_SESSION` even when `HERDR_SOCKET_PATH` is set and no cli `--session` was passed.

given the documented precedence, the exact socket override should win over `HERDR_SESSION` in that case. this command should use `/tmp/herdr.sock` and ignore the invalid session env:

```bash
HERDR_SOCKET_PATH=/tmp/herdr.sock HERDR_SESSION='bad/name' herdr workspace list
```

please skip `HERDR_SESSION` validation when `HERDR_SOCKET_PATH` is present and there is no explicit cli `--session`, and add a regression test for that precedence case.

src/session.rs
```diff
@@ -0,0 +1,483 @@
...
+
+    let mut line = String::new();
+    BufReader::new(stream)
+        .read_line(&mut line)
+        .map_err(|err| err.to_string())?;
+    let response: serde_json::Value = serde_json::from_str(&line).map_err(|err| err.to_string())?;
+    if let Some(error) = response.get("error") {
+        return Err(error.to_string());
+    }
+    wait_until_stopped(&socket_path, STOP_WAIT_TIMEOUT);
```
ogulcancelik: `wait_until_stopped()` has no return value, so `stop_session()` returns success even if the server is still running after the timeout. then `src/cli.rs` prints `"stopped": true` unconditionally.

please make the stop wait return a result/bool and fail `herdr session stop <name>` if the session is still reachable after the timeout. the command should not exit 0 or claim `"stopped": true` unless the socket is actually stopped.

ogulcancelik (changes requested):
thanks for the quick follow-up. overall this is very close to what i had in mind.

i’m requesting changes for two correctness issues before merge. with those two fixes, this looks good to me.

fbettag:
Thanks again for the review. I pushed a follow-up that addresses both requested correctness fixes:

- `HERDR_SOCKET_PATH` now wins over an inherited invalid `HERDR_SESSION` unless `--session` was explicitly passed.
- `session stop` now fails if the socket remains reachable after the stop timeout, instead of reporting success too early.

I also rebased the branch onto current `master`; GitHub now reports the PR as mergeable. Ready for another look.

ogulcancelik:
@fbettag ty! i'll fix the tests and merge

PR #61, opened 2026-04-27
fix: resize w claude code no flicker

### LINKED ISSUE #59: mouse-based pane resizing failed when panes had mouse reporting enabled (CLAUDE_CODE_NO_FLICKER)
# Summary 
The bug is triggered when Claude-code's [full screen mode](https://code.claude.com/docs/en/fullscreen) is enabled. The mouse-based-pane resizing does not work.

## Root cause

During `MouseDown` and `MouseDrag`, the app unconditionally forwarded mouse button events to the pane if the cursor was over it. This swallowed the resize-initiating events before `self.drag` could be set to a resize drag. Result: dragging a pane border had no effect — the split never moved.

##Fix
Gate pane mouse forwarding with `if self.drag.is_none()`:
- On `MouseDown` over a split border: `drag` stays `None`, so forwarding is skipped, resize drag starts normally
- On `MouseDrag` during a resize: `drag` is `Some(...)`, so forwarding is skipped, resize continues
- On normal clicks inside a pane: `drag` stays `None`, forwarding proceeds as before

EYH0602:
Fixes #59

### DIFF
--- a/src/app/input/mouse.rs
+++ b/src/app/input/mouse.rs
@@ -1215,8 +1215,8 @@ mod tests {
         assert_ne!(root_layout_ratio(&before), root_layout_ratio(&after));
     }
 
-    #[test]
-    fn dragging_vertical_pane_split_still_resizes_when_pane_mouse_reporting_is_enabled() {
+    #[tokio::test]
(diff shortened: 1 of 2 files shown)


ogulcancelik:
ty!

PR #84, opened 2026-05-09
Show focus on active agent in agents panel like spaces panel does

soomtong:
In herdr, there is a panel on the left that displays the workspace and agent status. 
I noticed that while the currently active item is clearly distinguished in the workspace list, active agents are not distinguished in the agents list. Therefore, I am submitting this change. 

- When there are no agent active in working session area
- When there are detected agents but no focus
- When there is a detected agent and it has focus

Tested in Ghostty and iTerm2

image

image

image

image

It is currently working well. 

Thank you for creating such a great tmux alternative. It’s incredibly useful and fits my needs perfectly.

### DIFF
--- a/src/app/state.rs
+++ b/src/app/state.rs
@@ -735,6 +735,29 @@ impl AppState {
             (24, 80)
         }
     }
+
+    /// Returns true when the given (workspace, tab, pane) refers to the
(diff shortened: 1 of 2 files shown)


ogulcancelik:
ty!

PR #91, opened 2026-05-10
chore: format html markups

soomtong:
It's a minor detail, but would it be okay to fix this? :)
Unless it was generated automatically, I think it would be fine to merge it. 

My editor uses a 120 char width to margin (on a standard QHD 27-inch monitor), and the readability isn't bad.


ogulcancelik:
i dont care too much about website formatting atm ty :)

PR #103, opened 2026-05-12
fix(remote): keep bridge socket path under sun_path ceiling

### LINKED ISSUE #102: herdr --remote fails on macOS with SUN_LEN error for ~24-char targets
On macOS, running `herdr --remote <target>` errors out before any SSH connection:

```
Error: Error { kind: InvalidInput, message: "path must be shorter than SUN_LEN" }
```

The local bridge socket is built as `{TMPDIR}/herdr-remote-{pid}-{target}-{session}.sock`. macOS' per-user TMPDIR is `/var/folders/.../T/` — about 49 chars on its own — so once the target is around 24 chars, the full path goes over `sun_path`'s 104-byte ceiling and `UnixListener::bind` rejects it. Linux doesn't hit this because `/tmp/` is only 5 chars and the limit is 108.

`sanitize_path_component` caps each piece at 32 chars, but that's per-component and doesn't account for the directory prefix, so it doesn't help here.

### Proposed fix
Pick the first form that fits under 103 bytes (one byte of headroom; portable across macOS' 104 and Linux's 108):

1. Readable `herdr-remote-{pid}-{target}-{session}.sock` in TMPDIR (unchanged when it fits — keeps debuggability).
2. Hashed short name `herdr-r-{pid}-{target-prefix}-{8byte-hash}.sock` in TMPDIR (covers the macOS overflow).
3. Same hashed name in `/tmp` as a last resort (for the rare case TMPDIR itself is longer than the budget).

Happy to send the PR once approved.

moonsphere:
Fixes #102. Approved in https://github.com/ogulcancelik/herdr/issues/102#issuecomment.

## Summary

`herdr --remote <target>` errors on macOS before SSH:

```
Error: Error { kind: InvalidInput, message: "path must be shorter than SUN_LEN" }
```

`local_forward_socket_path` built `{TMPDIR}/herdr-remote-{pid}-{target}-{session}.sock`. macOS' per-user TMPDIR (`/var/folders/.../T/`) is ~49 bytes, so a target around 24 chars pushes the full path past `sun_path`'s 104-byte ceiling and `UnixListener::bind` rejects it. The 32-char per-component sanitizer didn't help — it ignored the directory prefix.

## Fix

Three-tier fallback, picking the first form that fits under 103 bytes (one byte of headroom for the trailing NUL; portable across macOS' 104 and Linux's 108). Length is checked in bytes via `OsStrExt::as_bytes()`:

1. Readable `herdr-remote-{pid}-{target}-{session}.sock` in TMPDIR — unchanged when it fits, keeps grep-friendliness.
2. Hashed `herdr-r-{pid}-{target-prefix}-{8byte-hash}.sock` in TMPDIR — covers the macOS overflow case.
3. Same hashed name in `/tmp` — last resort when TMPDIR itself exceeds the budget.

The hash covers the **full unsanitized** target/session, so uniqueness doesn't depend on the prefix truncation; the prefix is kept only so the short-form filename remains debuggable.

## Tests

Three new cases under `remote::tests`:

- `local_forward_socket_path_uses_readable_name_when_it_fits` — short inputs keep the readable form.
- `local_forward_socket_path_fits_in_sun_path` — the original failing input (~24-char target) now fits via the hashed fallback.
- `local_forward_socket_path_falls_back_to_tmp_when_dir_is_long` — forces an over-long TMPDIR; asserts `/tmp` is used.

All three guard a shared mutex over `TMPDIR` mutations (same pattern as `api::tests::env_lock`).

End-to-end: `herdr --remote <target>` now reaches handshake; the socket file is cleaned up on `Drop`, no leftovers in TMPDIR/`/tmp`.

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -747,12 +747,50 @@ fn run_client_process(local_socket: &Path, reattach_command: &str) -> io::Result
 }
 
 fn local_forward_socket_path(target: &str, session_name: &str) -> PathBuf {
-    std::env::temp_dir().join(format!(
-        "herdr-remote-{}-{}-{}.sock",
-        std::process::id(),
-        sanitize_path_component(target),
-        sanitize_path_component(session_name)
-    ))
+    let pid = std::process::id();
+    let target_clean = sanitize_path_component(target);
+    let session_clean = sanitize_path_component(session_name);
+
+    let tmpdir = std::env::temp_dir();
+    let readable = tmpdir.join(format!(
+        "herdr-remote-{pid}-{target_clean}-{session_clean}.sock"
+    ));
+    if fits_unix_socket_path(&readable) {
+        return readable;
+    }
+
+    // macOS' per-user TMPDIR (~49 chars under /var/folders/...) can push the
+    // readable name past sun_path's 104-byte ceiling. Fall back to a hashed
+    // short name in TMPDIR, then to /tmp as a last resort when TMPDIR itself
+    // is longer than the budget. The hash covers the full unsanitized
+    // target/session so uniqueness does not depend on the prefix truncation;
+    // the prefix is kept only for debuggability.
+    let target_prefix: String = target_clean.chars().take(8).collect();
+    let hash = short_socket_hash(target, session_name);
+    let short_name = format!("herdr-r-{pid}-{target_prefix}-{hash}.sock");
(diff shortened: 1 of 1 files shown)


ogulcancelik:
lgtm ty!

Issue #128, opened 2026-05-16
Feature Request: long text selection with trackpads


leeeanh:
# Selection Autoscroll While Click-Dragging Text

> First-time contributor note: I intend to implement this myself and have a working
> branch ready. Requesting the `intends-to-pr` label. I'll wait for a `/approve`
> from a you before opening the PR.

## Current behavior

Click-drag selection can copy text within the visible pane, and dragging outside the pane currently scrolls once per drag event. However, when the pointer is held at the top or bottom edge, Herdr does not continue scrolling unless more mouse or wheel events arrive.

This makes long text selection difficult on trackpads, because users often cannot reliably keep dragging and also generate scroll events at the same time. Selections can also remain stale across some workspace, tab, or pane transitions.

## Requested change

Add Ghostty-style selection autoscroll while click-dragging text in a pane.

When an active text selection is being dragged near or beyond the pane edge, Herdr should continue scrolling the pane and extending the selection until the drag ends or the pointer returns to a safe area.

### Possible approaches

1. Add timer-based autoscroll state for active selections, following Ghostty's selection-scroll pattern.
2. Only improve wheel handling while a selection is active.
3. Add a keyboard-assisted selection mode for extending selections through scrollback.

Option 1 seems safest as the main fix because it addresses the root issue: Herdr needs follow-up scrolling even when no new drag or wheel events arrive.

### Triggering selection autoscroll should:

- Start from active drag-selection handling without changing plain click behavior
- Start when the pointer is in the top or bottom hot zone of the pane
- Start when the pointer is outside the pane boundary
- Preserve the existing immediate edge-scroll behavior for outside-boundary drag events
- Continue scrolling on a short timer while the drag remains active
- Extend the selection using the last known pointer position
- Stop on mouse release, copy, clear selection, key input, pane/layout invalidation, or when the pointer returns to the safe area
- Still copy the selected text normally on release

## How I plan to implement this

If the direction is approved, here is the concrete plan for the change:

- Start autoscroll when a dragging selection reaches the top/bottom pane hot zone, and keep the existing immediate edge-scroll behavior when dragging outside pane boundaries.
- Continue scrolling and extending the selection on a recurring ~30ms tick while the mouse is idle, using the last known pointer position.
- Stop autoscroll on MouseUp, key input, navigate actions, the cursor leaving the hot zone, hitting a scrollback boundary, or pane rect changes.
[...]

PR #129, opened 2026-05-16
feat: add selection autoscroll while dragging near pane edges

leeeanh:
@ogulcancelik this addresses the trackpad + keyboard long-text copy issue I ran into, described in #128.

If there's anything wrong with this MR, please let me know and I'll fix it.

---

## Current behavior

Dragging to select text near the top or bottom edge of a pane only scrolls when new drag or wheel events arrive. This makes long text selection unreliable on trackpads, where users often cannot keep holding a drag gesture while also generating scroll input.

Selections could also remain stale across some workspace, tab, or pane transitions.

## Changes

This adds Ghostty-style selection autoscroll for active text selection:

- Starts autoscroll when a dragging selection reaches the top/bottom pane hot zone
- Keeps the existing immediate edge-scroll behavior when dragging outside pane boundaries
- Continues scrolling and extending the selection on a recurring 30ms tick while the mouse is idle
- Stops autoscroll on MouseUp, key input, navigate actions, cursor leaving the hot zone, scrollback boundary, or pane rect changes
- Clears selection/autoscroll state across workspace, tab, and pane transitions to avoid stale selections

## Implementation notes

The feature uses Herdr's existing deadline-based scheduler instead of spawning a new timer task. `App` owns the autoscroll deadline, while `AppState` stores the active selection autoscroll state and pane rect snapshot.

Each tick validates that the selection, pane, runtime, and pane rect are still valid before scrolling. If any context changed, autoscroll stops conservatively.

## Testing

Added coverage for:

- Autoscroll deadline scheduling
- Self-healing when autoscroll state is cleared
- Stopping when pane rect changes
- Stopping when selection is cleared or finished
- Stopping at scrollback boundaries
- Hot zone and outside-boundary autoscroll state updates

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -113,6 +113,8 @@ impl AppState {
 
     pub fn switch_workspace(&mut self, idx: usize) {
         if idx < self.workspaces.len() {
+            self.selection = None;
+            self.selection_autoscroll = None;
             self.active = Some(idx);
             self.selected = idx;
             let workspace_id = self.workspaces[idx].id.clone();
@@ -196,6 +198,8 @@ impl AppState {
 
     pub fn switch_tab(&mut self, idx: usize) {
         if let Some(ws_idx) = self.active {
+            self.selection = None;
+            self.selection_autoscroll = None;
             let Some(ws) = self.workspaces.get_mut(ws_idx) else {
                 return;
             };
@@ -470,6 +474,8 @@ impl AppState {
         if self.workspaces.is_empty() {
             return;
         }
+        self.selection = None;
+        self.selection_autoscroll = None;
         self.mark_session_dirty();
         let terminal_ids = self.terminal_ids_for_workspace(self.selected);
         let workspace_id = self.workspaces[self.selected].id.clone();
@@ -590,6 +596,8 @@ impl AppState {
     }
 
     pub fn close_pane(&mut self) {
+        self.selection = None;
+        self.selection_autoscroll = None;
         self.mark_session_dirty();
         let active = self.active;
         let terminal_ids = active
@@ -615,6 +623,8 @@ impl AppState {
     }
 
     pub fn close_tab(&mut self) {
+        self.selection = None;
+        self.selection_autoscroll = None;
         self.mark_session_dirty();
         let should_close_workspace = self
             .active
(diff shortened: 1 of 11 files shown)


ogulcancelik:
thanks for the pr. the overall approach looks right, but i need a few changes before merge.

please update the branch on top of current `master`, then fix these:

1. run `cargo fmt`. current formatting fails `just check`.
2. stop selection autoscroll when `pane_scroll_metrics()` returns `None`. missing runtime/metrics is invalid context; it should stop autoscroll, not reschedule forever.
3. only continue timer autoscroll after the selection has actually become a drag. `is_in_progress()` includes anchored clicks, so an edge click or same-cell drag can currently start timer autoscroll.
4. update tests for those two behavior fixes. in particular, the no-runtime test should assert that autoscroll stops.
5. remove the stale `dead_code` / “phase 2” comments now that this code is active.
6. use `refs #128` in commit bodies if you keep issue references.
7. if you add or change user-facing docs, update `docs/next/`, not the root docs. no changelog update is needed.

after that i’ll rerun checks and review again.

leeeanh:
Thanks for your feedback, I will fix the issue base on your comments.

leeeanh:
Thanks for the detailed review, and apologies for not checking these more carefully before opening the PR. I've rebased onto current `master` and addressed the requested points:

  1. Ran `cargo fmt`; `just check` now passes locally.
  2. Selection autoscroll now stops when `pane_scroll_metrics()` returns `None`. Missing runtime/metrics is treated as invalid context and ends autoscroll instead of rescheduling.
  3. Timer autoscroll now only continues after the selection has actually become a drag, so anchored clicks and same-cell drags no longer keep it running. It no longer relies on `is_in_progress()` for this check.
  4. Updated the tests for both behavior fixes. The no-runtime test now asserts that autoscroll stops, and the scrollback-boundary tests now use an actually dragging selection.
  5. Removed the stale `dead_code` / "phase 2" comments now that the code path is active.
  6. Commit bodies now use `refs #128`.
  7. No user-facing docs changed, so nothing was added under `docs/next/` and no changelog entry was needed.

  Validation:
  - `just check` passes locally: 1015 nextest tests passed, plus 18 Python unittest checks passed.

  Branch is updated and ready for another look.

ogulcancelik:
ty!

Issue #140, opened 2026-05-17
Feature Request: Support 16-bit terminal themes


babymastodon:
### Current behavior

I just found Herdr on YouTube and tried it out - works great. So much better than my hobbled-togeter tmux config. Huge thanks for making this app.

Herdr ships several built-in themes, but there is no built-in theme that sticks to the terminal's own 16-color palette. If I want Herdr to blend with the rest of my terminal setup, I have to pick an RGB theme that may not match nearby TUIs.

### Requested change

Add a built-in `terminal` theme that uses only terminal colors instead of RGB values.

I asked codex to prototype this, and it doesn't look too bad. In the process, it was necessary to remove some of the RGB adjustments that were embedded in the UI code, since TUI themes must work with a fixed pallete. I think the cleaner direction is to move all surface contrast choices into `Palette` instead of doing color manipulation inside ui code.

Video: attached

https://github.com/user-attachments/assets/53e890d2-ec7d-42f0-ac2b-0c6b15758e0f

demo of the `terminal` theme cycling through a couple dozen Kitty themes. It looks decent across a wide range of palettes.

Prototype branch: <https://github.com/babymastodon/herdr/tree/terminal-theme>

### Reason

By using the built-in theme, terminals like Ghostty and Kitty can be configured to automatically follow your system dark/light theme, and works over SSH.

### Does this change UI, interaction, workflow, or product direction?

Not sure

### Impact details

Adding the `terminal` theme alone does not change the product direction. However, it might be worth considering `terminal` as the default, though I understand that is a weaker first-run experience if someone is on a poor or outdated terminal palette.

### Contribution intent

- [x] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

My branch is here, but not ready for PR (I haven't read the code yet). https://github.com/ogulcancelik/herdr/compare/master...babymastodon:terminal-theme?expand=1

Wanted to get alignment on the direction before I spend time cleaning it up. The above video is on kitty (linux gnome), but want to test it on ghostty mac a bit more since that's my daily work driver.

ogulcancelik:
thanks!

this direction makes sense and belongs in herdr. please go ahead with the pr.

/approve

[...]

PR #146, opened 2026-05-18
Support 16-bit terminal themes

babymastodon:
refs #140

Add a built-in terminal theme that uses only terminal colors instead of RGB values.

This also fixes a (presumed) bug where `dim_color` was crushing colors in navigation mode.

Video attached (with green accent color)

https://github.com/user-attachments/assets/68badf40-1a92-4d0d-a418-656b34fa4035

### DIFF
--- a/docs/next/CONFIGURATION.md
+++ b/docs/next/CONFIGURATION.md
@@ -216,7 +216,7 @@ command = "notify-send herdr 'custom command ran'"
 
 ## theme
 
-herdr ships with 17 built-in color themes. set one in config:
+herdr ships with 18 built-in color themes. set one in config:
 
 ```toml
 [theme]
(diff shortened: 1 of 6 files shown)


ogulcancelik:
code looks good to me.

could you change the pr body from `fixes #140` to `refs #140`, and amend the commit body to include:

refs #140

the commit body line is important because the release automation reads those `refs #...` lines, as mentioned in the contributing docs. i also avoid `fixes`/`closes` on normal prs because `master` contains unreleased work and github would close the issue too early.

babymastodon:
Done @ogulcancelik. Let me know if there's anything else needed for this PR.

ogulcancelik:
ty!

Issue #139, opened 2026-05-17
session.json writes break stow symlinks


cloudmanic:
### Current behavior

`session.json` is saved via `write(tmp) + rename(tmp, session.json)` in `src/persist/io.rs`. `rename` doesn't follow symlinks, so it replaces the symlink with a regular file.

### Expected behavior

If `session.json` is a symlink, the write lands on the target. The symlink stays a symlink.

### Reproduction

1. Replace `~/.config/herdr/session.json` with a symlink (e.g. `stow` from a dotfiles repo).
2. Launch herdr, trigger a save, quit.
3. `ls -l` — it's a regular file now.

### Impact

I stow `session.json` from my dotfiles repo. Every herdr run breaks the link and detaches the file from the repo.

### Environment

herdr 0.5.10, macOS, stowed from dotfiles.

### Contribution intent

- [x] I intend to open a PR.
- [x] I understand this does not mean the issue is approved.

/i-intend-to-pr

ogulcancelik:
this makes sense and belongs in herdr. please go ahead with the pr.

/approve

the bug is the current temp-write + rename path replacing the symlink itself instead of writing through to the target. please keep the fix narrow, preserve atomicity where practical, and add a regression test that covers session.json being a symlink.

PR #147, opened 2026-05-18
fix: preserve symlinks when saving session.json

cloudmanic:
`session.json` was written via tmp + atomic rename. `rename()` replaces the symlink at the destination with the new regular file, so any user who stows this file from a dotfiles repo loses the link on every save.

This change resolves symlinks before placing the `.tmp` so the rename targets the canonical file and the symlink at the input path is preserved. Atomicity is kept because the `.tmp` and the target stay in the same directory.

Three unit tests cover the symlink behavior: symlink to existing target, dangling symlink, and relative-target symlink. Verified end-to-end against a stow-like setup with a binary built from this branch.

refs #139


ogulcancelik:
ty!

Issue #199, opened 2026-05-19
Feat: Show named-session reattach command after detach


Golden-Pigeon:
### Current behavior

When I detach from a persistent Herdr session, the shell message says:

```
Run `herdr` to reattach
```

That is accurate for the default session, but it is not specific enough for named sessions.

For example, if I started or attached with:

```sh
herdr session attach work
```

then detaching still suggests herdr, which attaches to the default session instead of clearly pointing back to work.

### Requested change

When detaching from a named session, print the matching named-session attach command:

```
Run `herdr session attach work` to reattach
```

The default session should keep the current message:

```
Run `herdr` to reattach
```

The code change can stay small. For example, add a local helper in the client:

```rust
fn local_reattach_command() -> String {
    match crate::session::active_name() {
        Some(name) => format!("herdr session attach {name}"),
        None => "herdr".to_string(),
    }
}
```

Then use it in the local detach display path:

```rust
write!(f, "detached from server")?;
write!(f, "\nRun `{}` to reattach", local_reattach_command())?;
```

The existing remote branch should keep taking precedence through `HERDR_REATTACH_COMMAND`.

### Reason

Named sessions are useful when keeping multiple work contexts alive. After detach, the reattach hint should identify the same session the user just left, so the command can be copied directly without remembering the session name or accidentally opening the default session.

Besides, this fits Herdr’s persistence/session workflow and does not change the default behavior.

### Does this change UI, interaction, workflow, or product direction?

Yes

### Impact details

This only changes the informational text printed after detach. It does not add a new command, change keybindings, change session behavior, or affect default-session users.

For named sessions, it makes the existing detach/reattach workflow clearer.

### Contribution intent

- [x] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

Golden-Pigeon:
I noticed two similar places where the command hint may also be too generic for named sessions:

- Auto-detect compatibility errors currently suggest `herdr server stop` and then `herdr` again. For a named session, it would be clearer to suggest either `herdr session stop <name>` or `herdr --session <name> server stop`, then `herdr session attach <name>`.
- `herdr update` errors/prompts that require stopping a running server also suggest `herdr server stop`. If the active target is a named session, `herdr session stop <name>` or `herdr --session <name> server stop` would be more precise.

These seem like the same class of issue: user-facing recovery commands should point at the active session when one is known.

ogulcancelik:
/approve thanks, this direction looks good.

[...]

PR #234, opened 2026-05-21
feat: show named session command hints

Golden-Pigeon:
Refs #199.

## Summary

- centralize local session attach/stop command rendering
- show named-session guidance as `herdr session attach <name>` and `herdr session stop <name>`
- keep default-session guidance unchanged as `herdr` and `herdr server stop`
- preserve remote reattach command precedence over local session hints

## Behavior

Before this change, some runtime guidance still suggested default-session commands even when Herdr was running in a named session. That made detach/reattach or stop/retry instructions ambiguous for users with multiple persistent sessions.

With this change:

- default session reattach still says `herdr`
- named session reattach says `herdr session attach <name>`
- default session stop guidance still says `herdr server stop`
- named session stop guidance says `herdr session stop <name>`
- remote reattach continues to use the explicit remote reattach command when available

## Tests

Added regression coverage for:

- default-session reattach hints
- named-session reattach hints
- remote reattach hint precedence
- default and named local attach/stop command helpers
- named-session stop/attach guidance in server compatibility errors
- named-session stop guidance in update errors

## Validation

` just check`
all tests pass.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -154,7 +154,11 @@ impl std::fmt::Display for ClientError {
                             write!(f, "\nRun `{reattach_command}` to reattach")?;
                         } else {
                             write!(f, "detached from server")?;
-                            write!(f, "\nRun `herdr` to reattach")?;
+                            write!(
+                                f,
+                                "\nRun `{}` to reattach",
+                                crate::session::local_attach_command()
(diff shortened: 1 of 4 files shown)


ogulcancelik:
ty

Issue #232, opened 2026-05-21
feat: improve github copilot cli state detection heuristics


LaneBirmingham:
### Current behavior

Herdr detects the GitHub Copilot CLI process, but it does not fully support its interactive state transitions. It does not recognize the interactive selection menu (e.g., "Select an option  [Use arrows to move, type to filter]") or the execution confirmation prompts (e.g., "Are you sure you want to execute..."), leaving the agent marked as idle/unknown instead of blocked. It also does not recognize "Generating suggestions..." as a working state.

### Requested change

Improve the heuristics in `src/detect.rs` for GitHub Copilot CLI by:
- Detecting interactive selection menus and confirmation prompts as `Blocked`.
- Detecting the "Generating suggestions..." output as `Working`.
- Adding unit tests for these new cases.
- Promoting GitHub Copilot CLI in the supported agents tables in `README.md` and `docs/next/README.md`.

### Reason

GitHub Copilot CLI is a widely used tool. Bringing its state detection inline with other fully supported CLI agents ensures herdr users get precise, non-disruptive rollups and notifications when Copilot is waiting for input or working.

### Does this change UI, interaction, workflow, or product direction?

No

### Impact details

*No response*

### Contribution intent

- [x] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

/i-intend-to-pr

ogulcancelik:
/approve

thanks! you can go ahead with a pr.

i cannot test github copilot locally, so please make sure the change follows the existing default heuristic patterns.

one tip: you can run copilot in a separate pane, then use the main pane to ask your agent to read --ansi from idle, busy, and working states. that should help you verify the detector behavior.

LaneBirmingham:
@ogulcancelik Pull request ready for review https://github.com/ogulcancelik/herdr/pull/256

PR #240, opened 2026-05-21
feat: enhance github copilot cli state detection heuristics

LaneBirmingham:
refs #232


LaneBirmingham:
Closing while still testing to avoid CI triggers. Will Re open when more confidient with testing.

Issue #217, opened 2026-05-20
Persist pane screen history across server restarts


icedac:
### Current behavior

Herdr restores saved workspaces, tabs, panes, cwd, layout, focus, labels, and agent metadata after a server restart. It does not persist the actual pane screen history, so `herdr server stop` followed by a fresh start brings the panes back with fresh shells but without the output/scrollback that was visible before stopping the server.

### Requested change

Persist each live pane's terminal text history when the server saves its session during stop, then seed that history into the restored pane terminal when herdr starts again. This is similar in spirit to tmux-resurrect's optional pane-contents restore, but scoped to herdr's existing session restore path: restore the text/scrollback view, not the old process state.

### Reason

Herdr already owns the terminal backend and keeps scrollback while the server remains alive. Persisting that same pane history across intentional server stops makes restart behavior match the user's mental model: the workspace comes back with both its panes and the recent terminal context needed to continue work.

### Does this change UI, interaction, workflow, or product direction?

Yes.

### Impact details

This changes persistence behavior only. There is no new UI and no new command. Restored panes should still start fresh shells in their saved cwd; the saved history is replayed into herdr's terminal backend and must not be written to the PTY.

/i-intend-to-pr

---

https://github.com/user-attachments/assets/61db91b9-6971-41ee-abcf-37ede18e6c59

ogulcancelik:
/approve

this direction looks good to me. restore pane contents as view/history only, not process state, and don't write the saved text into the pty.

one requirement before pr review: please make pane-history persistence opt-in via a setting. pane output can include secrets, tokens, prompts, and command output, so the default should preserve today's behavior: restore workspaces/tabs/panes/cwd/layout/focus, but don't persist pane contents unless enabled.

tmux-resurrect handles this similarly: pane content capture is useful, but it is explicit opt-in.

please open the pr after the keybinding refactor lands. if i forget to follow up here, that refactor should land within the next 1–2 days.

PR #248, opened 2026-05-22
feat: persist pane screen history

icedac:
## Summary

- Persist opt-in pane screen history across full server restarts in `session-history.json`.
- Restore saved pane history as terminal view/history only, before the new shell starts, without writing saved text into the PTY.
- Add the Settings > Experiments toggle and `[experimental] pane_history = true` config persistence.
- Document the opt-in behavior and sensitivity of saved pane output in next-release docs.

## Why

Refs #217.

Herdr already restores workspaces, tabs, panes, cwd, layout, and focus after a full restart, but pane contents are blank because runtime process state is intentionally not restored. This preserves recent screen context as view history while keeping the default behavior unchanged unless users explicitly opt in.

Pane output can include secrets, tokens, prompts, and command output, so the setting defaults off and saved contents live separately from the structural session snapshot.

## Validation

- `cargo test --locked pane_history -- --test-threads=1`
- `cargo test --locked config::model::tests::experimental_config_parses -- --test-threads=1`
- `target/debug/herdr --default-config`
- `just check`

### DIFF
--- a/docs/next/README.md
+++ b/docs/next/README.md
@@ -79,7 +79,9 @@ tmux gives you persistence and panes, but it was built before agents existed. gu
 
 ## persistence
 
-start herdr where the work lives. locally, run `herdr`. it starts or attaches to the background session automatically, with no socket setup. run your agents, split panes, do your work. press `ctrl+b q` to detach. close your terminal, close your laptop; your agents keep running. open a new terminal, run `herdr`, you're back. same session, same panes, same agents.
+start herdr where the work lives. locally, run `herdr`. it starts or attaches to the background session automatically, with no socket setup. run your agents, split panes, do your work. press `ctrl+b q` to detach. close your terminal, close your laptop; your agents keep running. open a new terminal, run `herdr`, you're back. same session, same panes, same agents. if you stop the server and later start herdr again, restored panes bring back workspaces, tabs, cwd, layout, and focus.
+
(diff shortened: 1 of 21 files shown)


ogulcancelik:
sorry been busy to review this, but i'm also working on some interesting feature connected to this, so will review it soon. you dont need to try to make code up to date as im committing a lot, i can fix rebase conflicts

ogulcancelik:
nvm this actually helps for the stuff i want to build on top of, ty!

icedac:
> nvm this actually helps for the stuff i want to build on top of, ty!
 
Not sure if this is the kind of thing you’re building on top of this, but I was originally exploring a
follow-up where Herdr could gracefully shut down live Codex / Claude Code sessions, capture their
resume hint/session id if available, and then restart them on the next launch using the agent’s own
resume command.

For example, when Codex or Claude is shut down gracefully, it leaves a resume hint like:

> codex resume 019e5d5a-b269-79c0-b3c6-a75363xxxxx
> claude --resume 8a01814e-7fb9-4006-ba61-e3c47d90xxxx

I considered adding it to this PR, but split it out to keep the pane-history restore PR small. I’m not
100% sure it’s the right direction, but cmux has a similar flow and it felt useful.
  
  
Also, I’m really happy to contribute. I’m a heavy tmux user, and I had been thinking that we need a
tmux-like tool built for AI agents. I was even wondering whether I should build one myself and where
I’d start, so finding Herdr at that exact moment was amazing. It fits my niche almost perfectly.

PR #262, opened 2026-05-23
feat: support placeholders in [worktrees].directory for sibling layouts

### LINKED ISSUE #261: Support placeholders in [worktrees].directory for sibling-folder layouts
## Problem

`[worktrees].directory` is a single global root, so every repo's worktrees get filed under the same tree (`<directory>/<repo>/<branch-slug>`). 

I personally like to keep worktrees sitting in a subfolder right next to the repo (e.g. `~/code/foo`'s worktrees go under `~/code/foo.worktrees`), but currently the common convention `<repo>.worktrees/` as a sibling of the repo isn't expressible.

The current closest workaround is a shared root like `~/worktrees/<repo>/<branch>`, which gives you central storage but not a true sibling layout.

## Proposed change

Let `directory` interpolate three placeholders, substituted per source repo at worktree-creation time:

- `{repo_root}` — absolute path of the source repo, e.g. `/home/me/foo/bar`
- `{repo_parent}` — its parent directory, e.g. `/home/me/foo`
- `{repo_name}` — its basename, e.g. `bar`

Default behavior is unchanged: a literal `directory` (no placeholders) still resolves to `<directory>/<repo_name>/<branch-slug>` exactly as before.

Sibling-layout example:

\`\`\`toml
[worktrees]
directory = \"{repo_parent}/{repo_name}.worktrees\"
\`\`\`

A repo at `~/foo/bar` then gets checkouts at `~/foo/bar.worktrees/<branch-slug>`. When the template already encodes the repo name (`{repo_name}` or `{repo_root}` present), the implicit `<repo_name>/` segment is dropped so paths don't double up.

A couple of necessary caveats — happy to document:

- `{repo_root}`/`{repo_parent}` use Git's reported repo path, with symlinks resolved (since they come from `git rev-parse --show-toplevel`).
- `directory = \"{repo_root}/.worktrees\"` puts checkouts inside the working tree, so users will want to add the folder to `.gitignore`.

## Implementation

I have this implemented on a fork with tests and docs: https://github.com/ZainRizvi/herdr/tree/worktree-directory-template — happy to open a PR if this direction is acceptable. The change is reasonably contained (~5 files, ~280 lines mostly tests). It also adds startup + live-reload diagnostics for empty templates, unknown placeholders, and unclosed `{`, so typos surface immediately instead of silently becoming literal directory segments.

If you'd prefer a different API shape (e.g. a separate `layout = \"sibling\"` mode field rather than placeholders), I'd rather know before sending the PR.

---

/i-intend-to-pr

(Note: I opened #262 as a PR before reading CONTRIBUTING.md — apologies. It was auto-closed by the bot, which is the right behavior. The branch is still up at https://github.com/ZainRizvi/herdr/tree/worktree-directory-template as a candidate implementation for whatever direction you land on here.)

ZainRizvi:
Closes #261.

## Summary

Add `{repo_root}`, `{repo_parent}`, and `{repo_name}` placeholders to `[worktrees].directory` so worktree checkouts can live anywhere relative to each source repo — most usefully, in a sibling folder next to each repo:

\`\`\`toml
[worktrees]
directory = \"{repo_parent}/{repo_name}.worktrees\"
\`\`\`

A repo at `~/foo/bar` now gets worktrees at `~/foo/bar.worktrees/<branch-slug>`.

**Default behavior is unchanged.** A literal `directory` (no placeholders) still produces `<directory>/<repo>/<branch-slug>` exactly as before, so existing user configs see no path differences.

## How it works

- Placeholders are substituted per source repo at worktree-creation time (not at config load), so one `directory` value can serve repos under different parents.
- When the template includes `{repo_name}` or `{repo_root}`, the implicit `<repo_name>/` segment is dropped to avoid path doubling (e.g. `~/foo/bar.worktrees/<branch>`, not `~/foo/bar.worktrees/bar/<branch>`). `{repo_parent}` alone keeps the segment (it's treated like a legacy multi-repo root that happens to live next to the parent).
- `~` expansion runs after placeholder substitution, so `~/wt/{repo_name}` works.

## Diagnostics

Bad templates surface as user diagnostics at both startup and live config reload (mirrors the existing `ui.sound.diagnostics()` chain), rather than being silently treated as literal directory segments:

- Empty / whitespace `directory`
- Unknown placeholders, e.g. `{repo_nam}` typo — every typo in the template is reported, not just the first
- Unclosed `{`

The template still gets applied (warn-only) so the user sees the warning and can fix the config without restarting.

## Docs

Updated `docs/next/website/src/content/docs/configuration.mdx` with the placeholder list and two caveats (symlink resolution via `git rev-parse --show-toplevel`, and the `.gitignore` reminder when using `{repo_root}/.worktrees`).

## Test plan

- [x] `cargo test -- --test-threads=1` → 1330 passed, 0 failed.
- [x] `cargo clippy --tests -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [x] Manual: built locally, configured `directory = \"{repo_parent}/{repo_name}.worktrees\"`, created a worktree via `prefix+shift+g` — landed at `~/.../<repo>.worktrees/<slug>` as expected.

New tests cover, in `src/worktree.rs`:

- Legacy literal layout (`<root>/<repo>/<branch>`) unchanged.
- `{repo_parent}/{repo_name}.worktrees` produces sibling layout; implicit segment dropped.
- `{repo_root}/.worktrees` nests inside the repo; implicit segment dropped.
- `{repo_parent}/shared-worktrees` (no `{repo_name}`/`{repo_root}`) keeps the implicit `<repo>/` segment.
- `~` + placeholder combined (`~/wt/{repo_name}`).
- Placeholder inside literal segment (`prefix-{repo_name}-suffix`).
- Same placeholder repeated (`{repo_name}/old/{repo_name}`).
- Edge: repo at filesystem root (`/`) — no `//path`.
- Edge: repo at `/foo` whose parent is `/` — no `//path`. Asserts on `display().to_string()` rather than `PathBuf` equality (which normalizes `//` and would hide bugs).
- Diagnostics: empty/whitespace template, unknown placeholder (single + multi-typo), unclosed `{`, doubled `{{repo_name}}`, `{}` and `{ }`, accepts all known placeholders + verifies they're all actually substituted.

In `src/config/model.rs`:

- Chains `WorktreesConfig::diagnostics()` into `Config::collect_diagnostics()` and tests that wiring (so a refactor dropping the chain doesn't silently regress startup diagnostics).

In `src/app/mod.rs`:

- End-to-end `reload_config` integration test: typo'd template fed via TOML produces a `ConfigReloadStatus::Partial` report, the diagnostic reaches `state.config_diagnostic`, and the value still applies (warn-only).

## Notes for review

- I considered a separate `layout = \"sibling\"` mode field as an alternative API. Went with placeholders because they're strictly more expressive (any user-defined relative layout, not just sibling) and don't require a new enum to maintain. Easy to switch if you'd prefer the mode approach — the change is contained to `src/worktree.rs` and `src/config/model.rs`.
- The `worktree_directory` field on `AppState` changed type from `PathBuf` (resolved) to `String` (raw template), since per-repo resolution now happens at use time. Verified that no socket-API response or snapshot serializes this field, so the type change is purely internal.
- The new diagnostic surfacing at config reload follows the same shape as `ui.sound.diagnostics()` extension at `apply_live_config`.

Reference branch: https://github.com/ZainRizvi/herdr/tree/worktree-directory-template

### DIFF
--- a/docs/next/website/src/content/docs/configuration.mdx
+++ b/docs/next/website/src/content/docs/configuration.mdx
@@ -89,6 +89,12 @@ directory = "{repo_parent}/{repo_name}.worktrees"
 
 With this config, a repo at `~/foo/bar` gets worktrees at `~/foo/bar.worktrees/<branch-slug>`. When the template contains `{repo_name}` or `{repo_root}`, Herdr drops the implicit `<repo>/` segment so the path doesn't double up.
 
+A few caveats:
+
+- `{repo_root}` and `{repo_parent}` are the paths Git reports for the source repo, with symlinks resolved. If you open a symlink such as `~/foo/bar` that points at `~/work/projects/bar`, `{repo_parent}` will be `~/work/projects`, not `~/foo`.
+- `directory = "{repo_root}/.worktrees"` places checkouts inside the source repo's working tree. Add the chosen folder (e.g. `.worktrees/`) to `.gitignore` or `.git/info/exclude` so Git doesn't show the worktrees as untracked changes.
+- Unknown placeholders (e.g. `{repo_nam}` typos) and an empty `directory` are surfaced as config diagnostics on startup.
+
 Worktree actions are available from Git workspace rows. `New worktree` creates a branch and checkout, opens it as a new Herdr workspace, and groups it under the source workspace. `Open worktree...` lists existing Git worktree checkouts for that repo; choosing an already-open checkout focuses it, and choosing a closed checkout opens it in the same group.
 
 Grouped worktrees still behave like normal Herdr workspaces: they can be focused, renamed, closed, and contain their own tabs and panes. The parent row is the original workspace. Closing the parent row closes the whole Herdr group, but it does not delete checkout folders or branches.
--- a/docs/next/website/src/content/docs/configuration.mdx
+++ b/docs/next/website/src/content/docs/configuration.mdx
@@ -74,7 +74,20 @@ Set the root directory Herdr uses for Git worktree checkouts created from the si
 directory = "~/.herdr/worktrees"
 ```
 
-Herdr creates checkouts under `<directory>/<repo>/<branch-slug>`. For sibling-style checkouts, set this to a directory such as `~/Projects/herdr-worktrees`.
+By default Herdr creates checkouts under `<directory>/<repo>/<branch-slug>`. The `directory` value also accepts the following placeholders, substituted per source repo:
+
+- `{repo_root}` — absolute path of the source repo, e.g. `/home/me/foo/bar`
+- `{repo_parent}` — parent directory of the source repo, e.g. `/home/me/foo`
+- `{repo_name}` — basename of the source repo, e.g. `bar`
+
+For a true sibling-folder layout next to each repo, use `{repo_name}` in the template:
+
+```toml
+[worktrees]
+directory = "{repo_parent}/{repo_name}.worktrees"
+```
+
+With this config, a repo at `~/foo/bar` gets worktrees at `~/foo/bar.worktrees/<branch-slug>`. When the template contains `{repo_name}` or `{repo_root}`, Herdr drops the implicit `<repo>/` segment so the path doesn't double up.
 
 Worktree actions are available from Git workspace rows. `New worktree` creates a branch and checkout, opens it as a new Herdr workspace, and groups it under the source workspace. `Open worktree...` lists existing Git worktree checkouts for that repo; choosing an already-open checkout focuses it, and choosing a closed checkout opens it in the same group.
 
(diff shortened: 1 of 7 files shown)


ZainRizvi:
Understood — apologies for jumping straight to a PR. As a first-time contributor I should have followed the CONTRIBUTING.md flow: opened the issue with `/i-intend-to-pr`, waited for `/approve`, and skipped the next-release website docs mirror (I've now reverted that hunk on the branch).

Issue #261 has the proposal; happy to redo the PR if/when you `/approve` it. Leaving the implementation up at https://github.com/ZainRizvi/herdr/tree/worktree-directory-template as reference material if it helps the discussion.

Issue #65, opened 2026-05-04
Archive section for workspaces


leightonhughes:
### ✨ Add archive section for workspaces in sidebar

**Description**
It would be great to have an archive area in the sidebar to help manage and tidy up workspaces.

At the moment, the workspace list can get quite long. The only way to reduce clutter is to close workspaces, which makes it harder to revisit them later. If I want to return to a previous workspace, I have to create a new one instead.

**Expected behaviour**

* Ability to archive workspaces instead of closing/removing them
* Archived workspaces accessible in a dedicated “Archive” section in the sidebar
* Option to unarchive workspaces and restore them to the main list

**Actual behaviour**

* Workspaces must be closed to remove them from the list
* No way to easily revisit or restore previous workspaces

**Benefits**

* Keeps the sidebar clean and organised
* Makes it easy to revisit past work without recreating it
* Improves overall workflow efficiency

ogulcancelik:
i'll think about it, though i can't think about a good, intuitive and easy way to preserve archive for workspaces. will update this when i have ideas in mind.

PR #274, opened 2026-05-24
fix: raise AppEvent channel capacity to lift 64-pane ceiling

### LINKED ISSUE #265: Can't create more than 64 panes total across all workspaces and tabs
### Current behavior

Opening 8 workspaces with 2 tabs and 4 panes each lands at a total of 64 panes. Trying to create a 65th pane any such way fails to do anything. Can't open a new workspace. Can't open a new tab. Can't open a new pane.

If I close a pane, I can then open a new tab, but once I hit 64 total panes, I can not open any more.

I can't seem to find this behavior documented anywhere. It is unclear to me if this is a feature or a bug. Either way, the experience feels like an oversight considering what the tool is aiming to be useful for.

### Expected behavior

As an end user, I would expect 10 workspaces at 10 tabs each to not be unreasonable. Even with only 1 pane per tab, at 10x10, 100 panes is greater than 64.

### Reproduction

Open 8 workspaces, 2 tabs each, with 4 panes per tab. Try to open a 9th workspace, a 3rd tab, or a 5th pane, and it no-ops

### Impact

The impact is on technical and highly agentive workloads, which as I understand it is what this tool is supposed to be useful for. IMO, rich and deep workspaces are not unreasonable for herdr (correct if I am wrong):

I wouldn't necessarily expect all 100 tabs to have 16 panes open for the average use case, but I would expect there to be a bucket of expectable technical workloads that require a good chunk of panes:

An immediate one that comes to mind is monitoring `top` over a fleet of 50 or so virtual machines on an ultrawide or flat screen. In that case, you only have 14 panes left to do anything else, and that's assuming your fleet is limited to only 50 VMs _which is miniscule in production environments._

With this tool being used in agentive workloads, the other problem is that spawning 100+ agents is not unheard of, which this tool would not support monitoring for.

### Environment

herdr 0.6.0 on zsh on iterm2
Mac OS Tahoe 26.4.1
default config used.

### Contribution intent

- [ ] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

al3rez:
## Summary
- Bumps the bounded `AppEvent` mpsc channel in `src/app/mod.rs` from capacity **64** → **4096**, removing the hard total-pane wall reported in #265.
- Updates the backpressure regression test to fill the new capacity so it still exercises the same blocked-then-drained behavior.

## Why
Every pane spawned clones this `Sender` and runs blocking producer tasks against it. Once ~64 panes were live, steady-state event traffic could saturate the 64-slot buffer faster than the UI loop drained it, blocking both existing pane readers and the spawn path for pane #65 — manifesting as "can't open a new pane / tab / workspace" once the user hit 64 total panes (8×2×4 in the report).

Fixes #265.

## Test plan
- [x] `cargo test` — 1298 unit tests + integration suites pass
- [x] `full_internal_event_queue_eventually_applies_working_to_idle_transition` still verifies backpressure at the new capacity
- [ ] Manual: open >64 panes across multiple workspaces/tabs and confirm new panes/tabs/workspaces continue to spawn

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -221,7 +221,7 @@ impl App {
     ) -> Self {
         let (prefix_code, prefix_mods) = config.prefix_key();
         crate::kitty_graphics::set_enabled(config.experimental.kitty_graphics);
-        let (event_tx, event_rx) = mpsc::channel::<AppEvent>(64);
+        let (event_tx, event_rx) = mpsc::channel::<AppEvent>(4096);
         let render_notify = Arc::new(Notify::new());
         let render_dirty = Arc::new(AtomicBool::new(false));
 
@@ -2627,7 +2627,7 @@ mod tests {
             AgentState::Working
         );
 
-        for i in 0..64 {
+        for i in 0..4096 {
             app.event_tx
                 .try_send(AppEvent::UpdateReady {
                     version: format!("9.9.{i}"),


nick-the-nuke:
> Hi @al3rez, thanks for your interest in contributing!
> 
> We ask new contributors to open an issue first before submitting a PR. This helps us discuss the approach and avoid wasted effort.
> 
> Herdr is opinionated about how it should look, feel, and work. If your change affects UI, behavior, interaction patterns, or product direction, discuss it in an issue first.
> 
> **Next steps:**
> 
> 1. Open an issue describing what you want to change and why
> 2. Keep it concise and write in your own voice
> 3. If a maintainer approves with `/approve`, you will be added to the approved contributors list
> 4. Then you can submit your PR
> 
> This PR will be closed automatically. See https://github.com/ogulcancelik/herdr/blob/master/CONTRIBUTING.md for more details.

Bad bot. #265 is open for this

Issue #284, opened 2026-05-25
Skip unused libghostty-vt xcframework in Cargo builds


chenrui333:
## Current behavior

Building Herdr through `cargo install` asks the vendored `libghostty-vt` Zig build to emit `libghostty-vt` and, on macOS when `xcodebuild` is available, also emit a macOS xcframework.

The Rust build script only links `zig-out/lib/libghostty-vt.a`, so the xcframework artifact is not used by the Cargo build. In Homebrew source-build packaging, that extra `xcodebuild -create-xcframework` step can fail before the Rust binary links.

## Desired behavior

When `build.rs` invokes `zig build -Demit-lib-vt` for the Cargo build, also pass `-Demit-xcframework=false` so Cargo builds only the static library it links.

## Why this belongs in Herdr

This keeps source builds friendlier for Homebrew and other distro/package-manager environments without changing runtime behavior or user-facing workflows.

Reference: https://github.com/Homebrew/homebrew-core/pull/283413

/i-intend-to-pr

ogulcancelik:
/approve

i assume your intend to pr was not a mistake? well thanks, go ahead. let me know if you need something from my end

ogulcancelik:
ty will check it soon, pr gates having issues with branch rulesets atm

PR #286, opened 2026-05-25
build: skip unused libghostty-vt xcframework

chenrui333:
Summary:
- pass `-Demit-xcframework=false` when `build.rs` invokes the vendored `libghostty-vt` Zig build for Cargo
- avoid producing an unused macOS xcframework artifact during source package builds

Validation:
- `ZIG=/opt/homebrew/opt/zig@0.15/bin/zig just check`

Refs:
- #284
- Homebrew/homebrew-core#283413

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -60,7 +60,8 @@ fn main() {
         .arg(format!("-Doptimize={optimize}"))
         .arg(format!("-Dsimd={simd}"))
         .arg(format!("-Dtarget={zig_target}"))
-        .arg(format!("-Dversion-string={version_string}"));
+        .arg(format!("-Dversion-string={version_string}"))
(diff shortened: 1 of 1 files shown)


chenrui333:
@ogulcancelik thanks for the quick merge, should I expect 0.6.3 release, or we can just ship 0.6.2 on the homebrew side? :)

ogulcancelik:
@chenrui333 I think 0.6.2 is stable enough for Homebrew core. I’m releasing Herdr pretty frequently right now, but I don’t want to block the formula on every new patch release.

Hopefully future updates can be handled by the Homebrew update bots once this is in core. If anything does not get picked up automatically, I’m happy to send follow-up bump PRs.

Thanks for helping get this merged!

Issue #142, opened 2026-05-17
Feature Request:  Add tmux-style double-click word copy


babymastodon:
### Current behavior

Herdr already supports mouse drag selection and copy, but double-clicking terminal output does not select/copy the word or token under the cursor the way tmux does.

### Requested change

Add tmux-style double-click copy for pane output.

Expected behavior:

- Double-clicking a normal shell pane selects and copies the token under the cursor.
- The copied token should stay highlighted briefly, about 500ms, so there is visible feedback that it copied.
- It should have reasonable splitting heuristics to capture the most common tokens (filepaths, urls, variable names, etc), and ignore punctuation appropriately.

I asked codex to prototype this, and the video is attached:

https://github.com/user-attachments/assets/a0e773fd-b1d3-496d-a62d-d430242230ee

Prototype : https://github.com/babymastodon/herdr/compare/terminal-selection...babymastodon:herdr:double-click-word-select?expand=1

### Reason

I use this feature quite heavily in tmux, and noticed it within minutes of switching to herdr. It's probably not a super well known feature, but really convenient once you are in the know.

### Does this change UI, interaction, workflow, or product direction?

No

### Impact details

Does not change or conflict with any existing behavior

### Contribution intent

- [x] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

My branch is here, but not ready for PR: https://github.com/babymastodon/herdr/compare/terminal-selection...babymastodon:herdr:double-click-word-select?expand=1

Wanted to get alignment on product requirements before cleaning it up further.

babymastodon:
Hey @ogulcancelik  - i know you are super busy with tons of issues now that this repo is blowing up in popularity (congrats!)

Any objections if I move forward with the PR for this feature? It has 2 thumbs up from other people, so I think folks will find it useful.

No hurry at all. The prototype code is already done and I'm stress testing it this week at work. I can dedicate some time over the weekend to clean it up and ensure it meets code quality bar.

babymastodon:
Thanks! I'll send a PR this weekend.

babymastodon:
Thanks for your patience. I think the PR is good to go. Tried to keep the mouse event handling logic as simple as possible and found some seemingly dead code in the process.

Let me know if you have any feedback.

PR #296, opened 2026-05-25
Add tmux-style double-click word copy

babymastodon:
refs #142

Add tmux-style double-click copy for terminal pane output.

Double-clicking now copies the token under the cursor and highlights for 500ms. Token detection handles common shell output like paths, URLs, variables, package flags, CJK text, and surrounding punctuation.

Mouse events are still forwarded unchanged when the child terminal has mouse reporting enabled.

I also removed some unreachable code in the mouse-up handler.

Video Attached:

https://github.com/user-attachments/assets/93aaa067-804f-4fd5-a016-f7c87c3b8cf3

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -6,8 +6,10 @@ use tracing::{info, warn};
 use crate::detect::{Agent, AgentState};
 use crate::events::AppEvent;
 use crate::layout::{find_in_direction, NavDirection, PaneId};
+use crate::selection::Selection;
 use crate::terminal::{EffectiveStateChange, TerminalStateMutation};
 use crate::workspace::WorkspaceGitStatus;
+use unicode_width::UnicodeWidthChar;
 
 use super::state::{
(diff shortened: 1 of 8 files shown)


ogulcancelik:
@babymastodon thanks, this is super close to merge :)  can you patch two things before we merge?

first, clear the pending pane double-click candidate when a click turns into a drag or when selection mouse-up completes, so a drag-copy followed by a quick click cannot be misread as a double-click.

second, make URL trailing delimiter trimming balance-aware. valid urls like `https://en.wikipedia.org/wiki/Foo_(bar)` should keep the closing `)`; only unmatched trailing wrappers should be stripped.

after those small fixes i'll be happy to merge!

babymastodon:
Done: https://github.com/ogulcancelik/herdr/commit/3311e79d99a61b8530847d83aa68ecdf90c1f854

Wasn't sure if I should be squashing commits together or keeping them separate for easier review. Let me know what you prefer.

ogulcancelik:
lgtm ty !

Issue #304, opened 2026-05-26
feat: Improve agents sidebar context


icedac:
### Current behavior

Herdr shows agent state in the sidebar, but multiple running agents can still be hard to tell apart at a glance. The agents list does not make pane context prominent enough, so different agents can look similar even when they are working in different panes.

The sidebar layout is also mostly hardcoded. Adding pane names, working durations, custom status text, or alignment choices as fixed display rules makes the agents sidebar more useful, but it also turns each display preference into another code-level layout decision.

### Requested change

Make the agents sidebar easier to scan by showing clearer per-agent context, starting with pane names so agents can be distinguished more easily.

Also show how long an agent has been working, with blocked approval interruptions treated as part of the same working span when they directly interrupt active work.

Rather than hardcoding each new label, add a small sidebar configuration model so users can control which metadata appears, how it is split across rows, and where right-aligned sidebar content begins. Expose the common configuration in the TUI settings screen instead of requiring manual config edits for normal use.

<img width="1093" height="766" alt="Image" src="https://github.com/user-attachments/assets/f2baa54a-b265-4434-95f2-7c271dd26d8d" />

### Reason

Herdr is built around managing AI coding agents inside terminal workspaces. When several agents are running at once, the sidebar is the fastest way to understand what is happening without switching panes. Pane names and working durations make that view more operational: users can tell which agent is which and how long each one has been active.

The configurable layout part keeps this from becoming a pile of one-off hardcoded display choices. The default layout can stay familiar, while users who rely heavily on multiple agents can expose the context they need.

### Does this change UI, interaction, workflow, or product direction?

Yes.

### Impact details

This changes the sidebar UI and adds a settings surface for sidebar layout. The intended default behavior is to preserve the current appearance for users who do not customize anything.

The new settings should follow Herdr's existing TUI patterns: keyboard navigable, mouse friendly, live-previewed where useful, and consistent with the rest of the settings screen. The config format should be data-driven enough to support one, two, three, or more sidebar rows without special-casing a fixed two-line layout.

https://github.com/user-attachments/assets/cae93692-b3fa-4b35-991c-6d14dcfbaa80

ogulcancelik:
i'm refactoring agent status reporting and side bar, lets see if this is still needed after that.

PR #305, opened 2026-05-26
feat: Improve agents sidebar context

icedac:
Refs #304

Issue: https://github.com/ogulcancelik/herdr/issues/304

## Summary

This PR turns the sidebar from a fixed, hardcoded display into a configurable layout surface for spaces and agents.

The default layout stays the same, but users can now decide which sidebar metadata is visible, how it is split across rows, where right-aligned content begins, and which color preset each item uses. The same model is exposed both in `config.toml` and in the in-app Settings UI.

## Why

This started from a concrete agents-sidebar problem: when multiple agents are running, they are hard to tell apart because the sidebar did not show the pane name. The first goal was to add pane names so agents could be distinguished more easily.

While working on that, it also became useful to show how long each agent had been working. Once the agents sidebar had pane names, working durations, and more display choices, continuing to hardcode each layout decision made less sense.

So the scope grew: instead of only adding a couple of fixed labels, this PR makes the sidebar configurable from the TUI.

## High-Level Changes

- Adds `[ui.sidebar.spaces]` and `[ui.sidebar.agents]` config sections based on `lines = [...]`.
- Treats each sidebar segment as a configurable item with visibility, order, row placement, and color preset.
- Makes `right_alignment` and `custom_status` first-class sidebar items instead of special appended behavior.
- Adds a Settings > Sidebar screen with live previews for spaces and agents.
- Lets agent sidebar layouts render one, two, three, or more rows based on config.
- Moves agent working-duration ownership into terminal state so `working -> blocked -> working` keeps one continuous timer, `working -> idle -> blocked -> working` starts a fresh span, and `working -> blocked -> idle` clears the frozen duration.
- Hardens sidebar config parsing, explicit item omission, config-save failure handling, hook report sequencing, and display-width truncation based on PR review findings.

## Validation

- `cargo fmt --check`
- `cargo clippy --all-targets --locked -- -D warnings`
- `cargo nextest run --locked --status-level fail --final-status-level fail --failure-output final --success-output never`
- `python3 -m unittest scripts.test_changelog scripts.test_vendor_libghostty_vt`
- `just check` passes 1582/1582 plus maintenance unittest 30/30.

### DIFF
--- a/src/app/config_io.rs
+++ b/src/app/config_io.rs
@@ -1,4 +1,6 @@
 use super::App;
+use crate::app::state::{AgentViewItem, AgentViewPreferences, SpaceViewItem, SpaceViewPreferences};
+use crate::config::{AgentViewField, SpaceViewField, ViewColorPreset, ViewItem};
 
 impl App {
     pub(super) fn update_config_file<F>(&mut self, error_context: &str, update: F) -> bool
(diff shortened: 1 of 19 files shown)


ogulcancelik:
i'm refactoring a lot of stuff regarding statuses and how sidebar looks so gonna close this one for now. also please align with me on issue before opening prs.

PR #313, opened 2026-05-27
fix: report release when opencode process exits

modib:
## Problem
When opencode exits, its herdr plugin doesn't notify herdr — the pane stays stuck in whatever state was last reported (typically working or idle).

## Solution
Add a releaseSync() function that sends pane.release_agent via the Unix socket synchronously (best-effort) from process exit handlers:

- beforeExit — graceful shutdown
- SIGINT (Ctrl+C)
- SIGTERM (kill)
- SIGHUP (terminal close)
- SIGQUIT (core dump)

The socket connection is unrefed so it won't prevent the process from exiting.

## Testing
Tested locally with opencode inside a herdr pane. Verified:
- release is sent on Ctrl+C, closing the terminal tab, and normal exit
- After release, herdr correctly clears the agent from the pane sidebar
- No impact on existing working/blocked/idle state reporting during normal operation

refs #82

### DIFF
--- a/src/integration/assets/opencode/herdr-agent-state.js
+++ b/src/integration/assets/opencode/herdr-agent-state.js
@@ -2,7 +2,7 @@
 // managed by herdr; reinstalling or updating the integration overwrites this file.
 // add custom hooks/plugins beside this file instead of editing it.
 // HERDR_INTEGRATION_ID=opencode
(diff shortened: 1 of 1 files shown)


modib:
Converted to issue: #314

Issue #354, opened 2026-05-30
herdr --remote drops on idle: the SSH bridge has no keepalive


SunskyXH:
### Current behavior

`herdr --remote` tunnels the client↔server protocol over a long-lived `ssh` bridge that sets no keepalive. An idle network/NAT/firewall timeout silently drops the tunnel after a few minutes, and the client exits with `lost connection to server` even though the remote server and panes are still running.

### Requested change

Run the bridge ssh through a generated config that `Include`s the user's own ssh config first, then adds `ServerAliveInterval`/`ServerAliveCountMax` as a fallback. Because ssh uses the first value it finds, any keepalive the user set still wins (including `ServerAliveInterval 0` to disable it); herdr's values apply only when none is set. Opt out with `[remote] manage_ssh_config = false`.

### Reason

`herdr --remote` exists to leave a session attached to a remote dev box, so an unpredictable idle disconnect undermines the whole point. A keepalive fallback that defers to the user's ssh config makes it survive idle out of the box without overriding anyone's setup.

### Does this change UI, interaction, workflow, or product direction?

Yes

### Impact details

No UI or keybinding change. Remote sessions stop dropping on idle network timeouts, and a new opt-out key `[remote] manage_ssh_config` (default true) is added to config.

### Contribution intent

- [x] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

PR #355, opened 2026-05-30
feat: keep remote bridge alive with an ssh config keepalive fallback

SunskyXH:
## Problem

`herdr --remote` tunnels its client↔server protocol over a long-lived `ssh` bridge with no keepalive, so an idle network/NAT timeout silently drops it and the client exits with `lost connection to server` — even though the remote server and panes are still alive.

## Change

Run the bridge `ssh` through a generated config that `Include`s the user's own ssh config first, then adds `ServerAliveInterval`/`ServerAliveCountMax` as a fallback. Because ssh uses the first value it finds, any keepalive the user already set still wins (including an explicit `ServerAliveInterval 0` to disable it); herdr's values apply only when none is set. No wire-protocol or server change. Opt out with `[remote] manage_ssh_config = false`.

refs #354


ogulcancelik:
this report looks legitimate: `herdr --remote` currently runs the long-lived bridge over plain `ssh -T` with no keepalive, so idle nat/firewall drops are a real failure mode.


one small thing i’d like tightened before merge: in `SshStdioBridge::drop`, remove the generated temp ssh config directory after joining the bridge thread, not before. that avoids a possible shutdown race where the thread could still try to start a bridge connection using a config path that has just been deleted.

also, please don’t change `docs/next/CHANGELOG.md` in this pr. only update docs if there is a specific user-facing doc page that needs it, and please make sure your commits have refs # in body.

with those adjusted, i'm happy to merge.

ogulcancelik (changes requested):
one small thing i’d like tightened before merge: in SshStdioBridge::drop, remove the generated temp ssh config directory after joining the bridge thread, not before. that avoids a possible shutdown race where the thread could still try to start a bridge connection using a config path that has just been deleted.

also, please don’t change docs/next/CHANGELOG.md in this pr. only update docs if there is a specific user-facing doc page that needs it, and please make sure your commits have refs # in body.

with those adjusted, i'm happy to merge.

SunskyXH:
Thanks for the review, I have fixed points you mentioned. Please take another look when you have a moment.

ogulcancelik (approved):

ogulcancelik:
lgtm ty

Issue #243, opened 2026-05-21
Flag emojis and other multi-codepoint grapheme clusters render as blanks in pane output


juliogc:
### Current behavior

Pane output that contains flag emojis (e.g. 🇧🇷, 🇦🇷) or other multi-codepoint grapheme clusters (e.g. ZWJ family sequences like 👨‍👩‍👧) renders as a blank space inside herdr panes, even when the host terminal renders them correctly outside herdr.

In `src/pane/terminal.rs::ghostty_buffer_symbol_into`, the symbol scratch is filled with the grapheme cluster, then validated against the cell's expected width:

```rust
// src/pane/terminal.rs:1035-1046
let expected_width = match wide {
    crate::ghostty::CellWide::Wide => 2,
    crate::ghostty::CellWide::Narrow | crate::ghostty::CellWide::SpacerHead => 1,
    crate::ghostty::CellWide::SpacerTail => 0,
};
let actual_width = symbol_scratch.width();
if actual_width != expected_width
    && !(wide == crate::ghostty::CellWide::Narrow && actual_width == 2)
{
    symbol_scratch.clear();
    symbol_scratch.push_str(ghostty_blank_symbol_for_width(wide));
}
```

For a Regional Indicator pair like 🇧🇷 (`U+1F1E7 U+1F1F7`):

- Ghostty classifies the cell as `CellWide::Wide` → `expected_width = 2`.
- `unicode_width::UnicodeWidthStr::width()` sums per-codepoint widths → returns `4` (each RI codepoint is width 2 in isolation).
- The existing exception only covers `Narrow + actual_width == 2`, so this case falls through and the cluster is wiped to `"  "`.

Same path strips other multi-codepoint Wide clusters (ZWJ family/profession sequences, etc.).

### Expected behavior

When Ghostty classifies the cell as Wide and the symbol is a multi-codepoint grapheme cluster whose summed codepoint widths exceed 2, herdr should trust the terminal's cell classification and emit the cluster as-is — symmetric with the existing `Narrow && actual_width == 2` exception that already trusts the host for clusters like 💳.

Concretely, an additional escape hatch in the same conditional:

```rust
if actual_width != expected_width
    && !(wide == crate::ghostty::CellWide::Narrow && actual_width == 2)
    && !(wide == crate::ghostty::CellWide::Wide
         && actual_width > 2
         && symbol_scratch.chars().count() > 1)
{
    symbol_scratch.clear();
[...]

PR #406, opened 2026-06-01
fix: enable grapheme cluster mode on embedded panes

juliogc:
## Summary

Multi-codepoint grapheme clusters (flag emojis, ZWJ sequences, skin-tone modifiers) rendered as separated cells inside herdr panes. Root cause: libghostty-vt only clusters such sequences when DEC private mode 2027 is enabled (`vendor/libghostty-vt/src/terminal/Terminal.zig:337-340`), and herdr never enabled it for embedded pane terminals.

Real Ghostty sets `default_modes.grapheme_cluster = true` at terminal construction (`vendor/libghostty-vt/src/termio/Termio.zig:230`), but the `lib_vt` C API doesn't expose an equivalent default-mode setter. The minimal fix is to write `\x1b[?2027h` via the VT stream right after `Terminal::new`, wrapped in a named helper for grep-ability.

Diff is 3 files, +74 lines:
- `src/ghostty/mod.rs`: new `Terminal::enable_grapheme_cluster_mode` helper
- `src/pane.rs`: call it after `Terminal::new` at both production pane-init sites
- `src/pane/terminal.rs`: two `#[cfg(test)]` regression tests that round-trip `🇧🇷` and `👨‍👩‍👧` through the production `ghostty_buffer_symbol_into` path

## Caveat

A child app that sends `RIS` (`\x1bc`) or a soft reset reverts the mode to libghostty's default off. Fixing that properly needs a `set_default_mode` FFI on the libghostty-vt side; happy to follow up with an upstream patch if desired.

## Screenshots

Live repro on a Claude Code statusline running inside a herdr pane. Status bar shows a World Cup 2026 match with country flags around the score.

**Before — master / pre-fix (mode 2027 off):**

Screenshot 2026-06-01 at 17 27 50

**After — this PR (mode 2027 enabled at pane init):**

Screenshot 2026-06-01 at 15 58 36

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy --all-targets --locked -- -D warnings` passes
- [x] `cargo nextest run --locked` — both new tests pass; only failure is the pre-existing `live_server_holds_one_pty_master_fd_per_pane` test, which requires real `/dev/ptmx` allocation and is flaky in any Docker container (unrelated to this PR)
- [x] Verified live against herdr 0.6.6 via socket API: `printf '🇧🇷 end\n'` shows `🇧 🇷  end` on master (split RIs + buggy SpacerTail-as-space in text mode) and `🇧🇷 end` once mode 2027 is on
- [x] Built a local release binary and confirmed flags render in a fresh `--session flag-test` (see screenshots)

Local build / test had to run inside a Linux Docker container because zig 0.15.2 fails to link on macOS 26.5 (Tahoe) — `_bzero` / `_clock_gettime` undefined symbols against the new SDK's libSystem. CI's `macos-latest` is on an older macOS where 0.15.2 still works, so this should pass on the runner.

refs #243


ogulcancelik (changes requested):
thanks for the fix. i am requesting changes because we should use the native libghostty mode api here instead of manually writing the escape sequence.

i recently updated the vendor in [5bd21b7](https://github.com/ogulcancelik/herdr/commit/5bd21b7ed468b5ff4d1a5b2ed7e211f401eb138a) (`chore: update libghostty vendor`), and that brought in `ghostty_terminal_mode_set`, already wrapped as `Terminal::mode_set`. please change `enable_grapheme_cluster_mode` to call `mode_set(2027, true)` and return/propagate the result from the pane init paths.

the manual `\x1b[?2027h` write works, but it bypasses the typed ffi path and cannot report errors.

juliogc:
applied as requested. `Terminal::enable_grapheme_cluster_mode` now calls `mode_set(2027, true)` and the pane init paths propagate the `Result` via `io::Error::other` (matching `enable_kitty_graphics`):

```rust
// src/ghostty/mod.rs
/// Enable grapheme cluster mode (DEC private mode 2027). With this on,
/// libghostty-vt stores multi-codepoint grapheme clusters (flags, ZWJ
/// sequences, skin-tone modifiers) in a single cell instead of splitting
/// them across cells based on East Asian Width.
pub fn enable_grapheme_cluster_mode(&mut self) -> Result<(), Error> {
    self.mode_set(2027, true)
}
```

```rust
// src/pane.rs — both PaneRuntime::from_handoff_fd and ::spawn_command_builder
let mut terminal = crate::ghostty::Terminal::new(cols, rows, scrollback_limit_bytes)
    .map_err(|e| std::io::Error::other(e.to_string()))?;
terminal
    .enable_grapheme_cluster_mode()
    .map_err(|e| std::io::Error::other(e.to_string()))?;
```

The mode is set successfully — verified end-to-end against a local release build:

```text
# mode_get on a fresh Terminal
mode_get(2027) before enable = false
mode_get(2027) after  enable = true   ✓
```

However, the visible TUI behavior is **still broken**. Running the same `printf '🇧🇷 🇦🇷 🇺🇸 hello\n'` repro in a pane spawned by the patched binary shows blanks where flags should be. Probing the same pane via the socket API surfaces a divergence between two libghostty FFI paths after `5bd21b7`:

```text
herdr pane read <id> --source visible --ansi   ⚽ 2026-06-11 16:00 🇲🇽 x 🇿🇦   ← libghostty page read: clustered
herdr pane read <id> --source visible          ⚽ 2026-06-11 16:00 🇲  x 🇿     ← cell-iteration FFI: extras dropped
```

The TUI render path (`src/pane/terminal.rs::ghostty_buffer_symbol_into`) uses the cell-iteration FFI via `RowCellIter::grapheme_text_into`, so it inherits the dropped-extras behavior and then falls back to the wide-blank symbol on the width-mismatch check — which is what the user sees.

Verified inline with a synthetic case under the same writer/reader pair:

| Method | Result for `printf '🇧🇷'` |
|---|---|
| `Terminal::read_text_viewport` (page-level) | `"🇧🇷"` (8 bytes — clustered ✓) |
| `RowCellIter::grapheme_text()` on the head Wide cell | `"🇧"` (4 bytes — extras lost ✗) |

Looking at `vendor/libghostty-vt/src/terminal/c/render.zig::rowCellsGetGraphemesUtf8` and the `cells.graphemes` field populated in `src/terminal/render.zig:520-534` (only fills `cells_grapheme[x]` when `page_cell.content_tag == .codepoint_grapheme`), I suspect the render snapshot isn't seeing the post-`appendGrapheme` tag for that cell — but I haven't bisected the vendored change closely enough to point at the exact regression.

Happy to either:
1. Land this PR as-is since it implements the requested mechanism, and follow up separately on the vendor cell-iteration regression
2. Hold this PR until the vendor cell-iteration path is reconciled, since the user-visible behavior still doesn't change without it

Whichever you'd prefer.

ogulcancelik:
@juliogc i think you forgot to push
https://github.com/ogulcancelik/herdr/pull/406/changes#diff-e731964f09ddd7ce673363a5ad9dfec5db2fac7987aa216e1ca78793bd2efb60R588-R589
still calls`self.write(b"\x1b[?2027h");` :)

ogulcancelik:
i have to close this as latest libghostty created some problems i'll handle it later on

juliogc:
> @juliogc i think you forgot to push [#406 (changes)](https://github.com/ogulcancelik/herdr/pull/406/changes#diff-e731964f09ddd7ce673363a5ad9dfec5db2fac7987aa216e1ca78793bd2efb60R588-R589) still calls`self.write(b"\x1b[?2027h");` :)

I didn't forget, actually. I've just decided to keep the changes locally while I doesn't have a clear direction to move on.
Sorry if the changes weren't very clear.

I'll commit it in my fork branch anyways to help keep this on-track.

---

> i have to close this as latest libghostty created some problems i'll handle it later on

Fully understandable, don't need to rush because of my cosmetic things.
Let me know if I can help with anything about it.

Issue #398, opened 2026-06-01
leaving copy mode should restore scroll position


reobin:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

when you leave copy mode, viewport stays exactly where you're at, forcing you to use some other way (mouse) to scroll back all the way down, or enter copy mode again.

### Expected behavior

just like tmux, when you leave copy mode, you should be sent back to exactly where you were before entering copy mode

### Reproduction

- enter copy mode with `prefix+[`
- scroll up a bunch (e.g. 3x `ctrl+u`)
- leave copy mode with `q`

### Impact

leaving copy mode often leaves me in a place where I need to enter copy mode again or use the mouse, which defeats the purpose of copy mode.

### Environment

- Herdr version: 0.6.5
- Operating system: arch linux
- Terminal: ghostty
- Shell, if relevant: bash

PR #410, opened 2026-06-02
fix: return to bottom after copy mode exit

reobin:
- will scroll all the way down on quit copy mode, except if the user scrolled up before entering copy mode. in that case, it will restore scroll to that position.

refs #398


ogulcancelik:
lgtm ty, next time dont forget to add "refs" to commit body please!

Issue #400, opened 2026-06-01
Auto-switch to an ASCII input source while prefix mode is active (macOS, CJK IME)


ogulcancelik:
### Discussed in https://github.com/ogulcancelik/herdr/discussions/380

<div type='discussions-op-text'>

<sup>Originally posted by **sf-jin-ku** June  1, 2026</sup>
### idea / problem

I use a Korean IME most of the day, and prefix commands are painful because of it.

When I'm typing Korean in a pane and hit `ctrl+b`, the prefix key gets through fine, but the next key (`c`, `v`, `h/j/k/l`, …) often gets eaten by the IME and composed as a Korean character instead of being read as a command. So I either have to switch to English first, or the command just doesn't land. This happens with any non-Latin IME, not just Korean.

### requested change

While prefix mode is active, Herdr briefly switches the system input source to an ASCII-capable layout, runs the command, then restores my IME to whatever it was when prefix mode ends. Pane text input stays exactly as it is now — only the prefix command layer changes.

I'd keep it small to start: macOS only, prefix mode only, best-effort (no-op if switching fails), restore on exit/cancel. On macOS this maps cleanly to the Text Input Source APIs — save the current source, select the current ASCII-capable layout, restore on exit — so it doesn't hard-code ABC/US, it uses whatever ASCII layout the user already has.

### why you want this

For anyone on a non-Latin IME, prefix commands feel unreliable right now — the command layer should be as predictable as it is in English input mode, without touching how panes handle text.

It's also not a new idea. The modal-editor world has done this for years (command mode = ASCII, text mode = keep my IME):

- vim-barbaric — switches input method when you leave insert mode (https://github.com/rlue/vim-barbaric)
- vim-xkbswitch — switches layout on insert enter/leave, restores on the way back (https://github.com/lyokha/vim-xkbswitch)
- VSCodeVim's `vim.autoSwitchInputMethod` — drops the IME when you exit insert mode (https://github.com/VSCodeVim/Vim)
- im-select / macism — the CLI switchers those plugins lean on (https://github.com/daipeihust/im-select, https://github.com/laishulu/macism)

Happy to work on it if the direction sounds good.</div>

ogulcancelik:
/approve @sf-jin-ku

thanks for the suggestion and appreciate the help on ime stuff. this direction makes sense, please go ahead with a pr.

please keep the first version narrow:

[...]

PR #434, opened 2026-06-03
feat: switch input source during prefix mode (macOS, opt-in)

sf-jin-ku:
## What this does

Implements #400. While prefix mode is active on macOS, switch the system input source to the current ASCII-capable layout, then put back whatever was active once prefix mode ends. Off by default, behind `[experimental] switch_ascii_input_source_in_prefix`.

The problem: with a Korean (or any non-Latin) IME active, hitting the prefix key works, but the next key often gets composed by the IME instead of read as a command. So prefix commands are unreliable unless you switch to English first.

## How it works

The switch hangs off the `Mode` transition rather than any new state. After each event is handled, I compare the previous mode against the current one: entering prefix switches to ASCII (only when the flag is on), and leaving prefix — for any reason, including cancel — restores. Keeping it on the mode edge means there's no second flag that can drift out of sync, and every exit path restores through the same point.

`AppState` only carries the bool; the actual macOS call lives in the runtime, so state stays pure. The macOS side is all in `src/platform/macos.rs` behind a small `PrefixInputSource` trait, so the rest of the app never sees the Carbon/CoreFoundation FFI and non-macOS just compiles a no-op. The saved source is held in a struct whose `Drop` does the restore + `CFRelease`, so it can't leak on an early return.

It uses `TISCopyCurrentASCIICapableKeyboardLayoutInputSource`, so it picks up whatever ASCII layout you already have — no hard-coded ABC/US. If the current source is already ASCII-capable (compared with `CFEqual`) or any TIS call fails, it does nothing and input keeps working.

## Testing

`PrefixInputSource` is a trait so the transitions are covered with a fake — enter/leave, flag off, and the already-ASCII / failed-switch case where the later restore has to stay harmless. There's also a dispatch-level test that Esc out of prefix restores, since the exit is decided below `handle_key`.

`just ci` passes; I also ran the wider `just check` (clippy `-D warnings` + the python unittests). Verified by hand on macOS with a Korean IME: prefix commands land mid-composition, and my IME comes back when prefix mode ends.

## Docs

New `[experimental]` option, so I updated `docs/next/website/src/content/docs/configuration.mdx` for the next release. Left `docs/next/CHANGELOG.md` alone (maintainers handle that at release), and didn't touch root README/CHANGELOG or the released website docs.


ogulcancelik:
lgtm ty!

Issue #337, opened 2026-05-28
Add support for Mastracode integration


wardpeet:
### Current behavior

If you use opencode, claude code etc you get a nice signal in the agents pane to see if something is busy or not. I would like to have this for mastracode as well.

https://code.mastra.ai/

### Requested change

Add a new integration for mastracode

### Reason

Makes me more productive

### Does this change UI, interaction, workflow, or product direction?

No

### Contribution intent

- [x] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

ogulcancelik:
/approve

i'll not able to verify the integration but i see that you are working with mastra so the support depends on you and i'll forward any issues regarding mastracode to you :)

wardpeet:
Thank you @ogulcancelik, yes you can always tag me!

PR #440, opened 2026-06-03
feat: add mastracode integration

wardpeet:
## Summary
- Add MastraCode as an installable Herdr integration target.
- Install a MastraCode hook asset into the MastraCode config directory.
- Wire MastraCode CLI parsing, integration recommendations/status, install, uninstall, and tests.

Refs #337

## Test plan
- `git diff --check`
- `sh -n src/integration/assets/mastracode/herdr-agent-state.sh`
- `HERDR_ENV=0 sh src/integration/assets/mastracode/herdr-agent-state.sh working <<<'{}'`

`just` and `cargo` are not available in this local environment, so the full Rust check suite could not be run here.

### DIFF
--- a/src/api/schema.rs
+++ b/src/api/schema.rs
@@ -511,6 +511,7 @@ pub enum IntegrationTarget {
     Opencode,
     Hermes,
     Qodercli,
+    Mastracode,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
--- a/src/cli/integration.rs
+++ b/src/cli/integration.rs
@@ -103,13 +103,13 @@ fn parse_integration_target(
 ) -> std::io::Result<Option<IntegrationTarget>> {
     let Some(target) = args.first().map(|arg| arg.as_str()) else {
         eprintln!(
-            "usage: herdr integration {action} <pi|omp|claude|codex|opencode|hermes|qodercli>"
+            "usage: herdr integration {action} <pi|omp|claude|codex|opencode|hermes|qodercli|mastracode>"
         );
         return Ok(None);
     };
     if args.len() != 1 {
         eprintln!(
-            "usage: herdr integration {action} <pi|omp|claude|codex|opencode|hermes|qodercli>"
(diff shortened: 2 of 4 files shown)


ogulcancelik:
@wardpeet thanks for the pr. the upstream conflict looks small enough to resolve; it is mostly newer integrations landing in the same registry code.

i checked mastracode hooks docs. the available events appear to be `sessionstart`, `userpromptsubmit`, `pretooluse`, `posttooluse`, `stop`, `sessionend`, and `notification`: https://code.mastra.ai/configuration#hook-events

my main question is whether those hooks are lifecycle-complete enough to use as the source of truth. mastracode has permission rules with `ask`, so approval prompts exist, but i do not see a dedicated permission/request hook. does `notification` fire for those approval prompts, or is there another hook payload we can use to reliably report `blocked`?

also, have you confirmed esc/interruption behavior? the docs say `stop` fires when an agent response ends as `complete`, `aborted`, or `error`, so i expect esc to produce `stop`, but it would be good to verify against the actual cli.

if the hooks are only event notifications rather than a complete status lifecycle, i would prefer not to make them the primary integration path. we had similar issues with claude code and codex: missed or ambiguous hook events made state drift, and we ended up relying on screen reads because herdr needs current visible status, not just event history. if mastracode has stable visible ui states, please consider adding screen heuristics instead; `src/detect/agents/codex.rs` and `src/detect/agents/claude_code.rs` are good references.

ogulcancelik:
@wardpeet i assume you don't have much time to support this. let me know when you want to :)

wardpeet:
Hey sorry, pushing a new feature. Afterwards i'll re-open. I've been using herdr non stop! :D

ogulcancelik:
> Hey sorry, pushing a new feature. Afterwards i'll re-open. I've been using herdr non stop! :D

let me know when you are able to support the pr :D also please check docs now herdr has two tiers of agents, first citizen agents who have full lifecylce hook supports or not, both require new manifest shapes, also agent detect cli should help you to debug the correct manifest regarding screens. if mastracode uses osc for titles even better!

Issue #459, opened 2026-06-04
copy mode page scrolling ignores tmux-style boundaries


reobin:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

In copy mode, Ctrl-U and Ctrl-D scroll by the visible pane height and do not move the cursor when scrolling clamps at the top of history or live bottom.

### Expected behavior

Page scrolling should match tmux-style copy mode: full-page movement leaves room for the prompt/status area, and any leftover movement at scrollback boundaries moves the copy cursor.

### Reproduction

  1. Open a pane with enough scrollback.
  2. Enter copy mode.
  3. Use Ctrl-U, or Ctrl-D near the top of history or live bottom.
  4. Notice that the viewport clamps but the copy cursor does not consume the remaining page movement, so it
  stays in the wrong row.

### Impact

  Copy mode navigation feels off and can leave the cursor in the wrong place at scrollback boundaries.

### Environment

  - Herdr version: 0.6.8
  - Update channel (stable or preview): stable
  - Operating system: archlinux
  - Terminal: ghostty
  - Shell, if relevant: bash

PR #460, opened 2026-06-04
fix: match copy mode page scrolling boundaries

reobin:
refs #459

* match copy mode page scrolling to tmux-style page sizing
* move the copy cursor when scrollback clamps at history top or live bottom
* add focused copy mode page-scroll coverage


ogulcancelik:
lgtm ty

Issue #560, opened 2026-06-11
Massive lag and high CPU usage in v0.6.9


tinysnake:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

I‘m using herdr in Windows 11 WSL debian trixie distro, and runs pi-coding-agent in herdr.

Every time I finish a session and create a new session, it just freezes. and I can see herdr sever using 100% cpu resource in `top` command. after about 1 min it will unfreeze.

Oh, I have pi-fff installed, and pi-fff refreshs file caches on session start, normally it just freezes a fraction of second.

And, `select text auto copy` function also broken in this version.

### Expected behavior

herdr  doesn't freeze.  `select text auto copy` works.

### Reproduction

Install pi-fff to pi-coding-agent and load a relatively large repo in pi.
Every time you create a new session, it freezes.

### Impact

The whole herdr freezes, cannot click , cannot response to any keyboard event.

### Environment

- Herdr version: 0.6.9
- Update channel (stable or preview): stable
- Operating system: Windows 11 WSL debian distro.
- Terminal: Windows Terminal
- Shell, if relevant: zsh
- Relevant config, if any: no.

williamleong:
I'm experiencing the same with the herdr --remote

Just running pi (0.79.1) increases CPU usage for herdr server from less than 1% to 20-50% in htop across 3 cores.

To add on, CPU usage only starts going up when I start pi and drops back to normal when I exit pi, so I guess it's something to do with herdr's pi integration. herdr has been pretty great for me so far, happy to see it improve.

- Herdr version: 0.6.9
- Update channel (stable or preview): stable
- Operating system: Ubuntu 24.04 6.17.0-29-generic (remote)
- Terminal: ghostty
- Shell, if relevant: bash
- Relevant config, if any: no.

sysilion:
I am also experiencing this high CPU issue (~200% on macOS).

Upon investigating the logs at ~/.config/herdr/herdr-server.log, I found an event spam loop:
```
2026-06-11T07:52:18.922044Z  INFO herdr::pane: agent changed pane=11 previous_agent=None agent=Some(OpenCode) process=opencode pgid=Some(78831)
```
This repeats thousands of times per second. It appears to be an event storm triggered when multiple opencode agents are active (I have 5 terminal sessions). Throttling or rate-limiting these state synchronization events seems like it would solve the problem.

[...]

PR #565, opened 2026-06-11
fix: stop lifecycle authority detection thrash

dzevs:
Refs #560.

## Summary
- make lifecycle-authority sync idempotent so repeated `true -> true` updates do not reset agent detection
- skip routine pty/process probing while full lifecycle hook authority is active
- preserve pending process-exit and release/reacquire probes
- add regression coverage for lifecycle-authority probe gating and reset notification behavior

## Root cause
For full lifecycle hook authority panes, app event handling repeatedly synced lifecycle authority into the runtime. The runtime setter notified the detection reset path on every `true` call, even when already active. That reset `AgentDetectionPresence` back to `None`.

The pty-first process probe then ran before the lifecycle-authority screen-detection gate and re-detected the foreground Pi process, emitting repeated `None -> Some(Pi)` state changes. Those state changes fed back into app event handling, creating the CPU/log/event-loop flood.

## Validation
- `cargo fmt --check` ✅
- `cargo test pane::tests:: -- --nocapture` ✅
- `cargo test --bin herdr` ✅ (`1879 passed`)
- `cargo test` ✅ (unit + integration tests)
- `cargo clippy --all-targets --locked -- -D warnings` ✅

## Manual verification
Tested locally with a rebuilt/installed `herdr 0.6.9` binary:
- running server binary hash matched the fixed build
- no renewed `previous_agent=None agent=Some(Pi)` flood after restart
- UI lag disappeared
- select-to-copy worked again

### DIFF
--- a/src/pane.rs
+++ b/src/pane.rs
@@ -263,6 +263,14 @@ fn foreground_group_changed(
         && (foreground_pgid.is_some() || last_foreground_pgid.is_some())
 }
 
+fn should_skip_process_probe_for_lifecycle_authority(
+    full_lifecycle_authority_active: bool,
+    process_exit_pending: bool,
+    release_pending: bool,
+) -> bool {
+    full_lifecycle_authority_active && !process_exit_pending && !release_pending
+}
+
 fn should_probe_foreground_job(input: ProcessProbeInput) -> bool {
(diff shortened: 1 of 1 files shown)


ogulcancelik:
lgtm correct fix ty

Issue #606, opened 2026-06-14
[idea] Support Devin CLI as a first-class agent in Herdr


ogulcancelik:
### Discussed in https://github.com/ogulcancelik/herdr/discussions/446

<div type='discussions-op-text'>

<sup>Originally posted by **minatoaquaMK2** June  3, 2026</sup>
### idea / problem

I use Devin CLI locally as a coding agent, but Herdr does not currently recognize it as a first-class agent.

That makes Devin sessions awkward to manage in Herdr compared with agents like Claude Code, Codex, Copilot CLI, or OpenCode. In practice, Herdr cannot reliably identify Devin panes, infer Devin-specific blocked / working states, or resume a Devin session after a Herdr restart.

### requested change

I’d like Herdr to support Devin CLI as a first-class agent.

Concretely, that would mean:

detect devin as a known agent process
infer Devin terminal states like idle / working / blocked from its UI
optionally support a Herdr-managed Devin integration hook for more reliable semantic state reporting
support native session restore with devin --resume <id>
Official docs:

https://docs.devin.ai/zh/work-with-devin/devin-cli
https://docs.devin.ai/

### why you want this

I use Herdr to coordinate multiple coding agents in parallel, and Devin CLI fits the same workflow as the agents Herdr already supports.

If Herdr understood Devin better, I could:

see Devin sessions in the same blocked / working / idle workflow as other agents
notice permission prompts or trust prompts without polling the pane manually
recover Devin sessions more cleanly after restarting Herdr
use Devin alongside Claude Code / Codex / Copilot in a more consistent way</div>

minatoaquaMK2:
I'd be happy to implement this. If the proposed direction looks good, could I get approval for a first-time contribution?

GitHub: @minatoaquaMK2

ogulcancelik:
/approve @minatoaquaMK2

thanks.

process detection for `devin` should be straightforward. the important part is state quality. please look at the existing agent manifests first, especially the patterns around osc-backed detection and fallback screen heuristics.

the best paths are:

1. full lifecycle authority through devin hooks, if devin can reliably report working/idle/blocked, session id, and release/exit behavior. kimi, hermes, and opencode are good references.
2. osc title/progress markers, if devin emits stable terminal title/status sequences like claude code and codex.
3. manifest-based screen heuristics as fallback, based on real `herdr agent read --source detection` captures.

[...]

PR #622, opened 2026-06-15
feat: add devin cli detection and restore support

minatoaquaMK2:
## Summary

  - Add Devin CLI process detection and native restore with `devin --resume <id>`.
  - Add `herdr integration install devin` hooks for session identity, lifecycle state
  reports, and release on session end.
  - Add Devin manifest fallback rules for idle, working, and blocked states when hooks
  are not installed or not active.
  - Update integration docs, config/schema listings, sound settings, resume support, and
  regression tests.

  ## State quality notes

  - `SessionStart` is session-identity-only. It reports the Devin session id through
  `pane.report_agent_session` and does not claim lifecycle authority.
  - Lifecycle authority starts only from state-producing hooks: prompt submission and
  tool-use report `working`, permission prompts report `blocked`, `Stop` reports `idle`,
  and `SessionEnd` releases the agent.
  - Verified Devin CLI `v2026.5.26-8` raw PTY OSC output. Devin emits OSC title entries,
  but they contain cwd/prompt/task title only, not stable idle/working/blocked markers,
  so this PR does not add OSC state rules.
  - Fallback screen rules are scoped to detection-source bottom-buffer regions and
  visible controls, not whole-pane incidental text.
  - The idle fallback includes a real `herdr agent read --source detection` capture from
  a no-hook Devin pane.

  ## Testing

  - `just check`

refs #606


ogulcancelik:
thanks, this is a strong first pass. the osc choice makes sense if the raw pty capture only showed cwd/prompt/task title and no stable lifecycle markers.

the main thing i want verified before treating `herdr:devin` as full lifecycle authority is the negative/interrupt path. since full lifecycle authority pauses screen detection, a missed hook transition can leave the pane stuck in `working` or `blocked`.

please verify the actual hook event sequence for these cases and either document it in the pr or adjust the integration:

- normal turn: `UserPromptSubmit` -> tool hooks if any -> `Stop` once the ui is idle
- permission approve: `PermissionRequest` -> a later hook that moves back to `working`
- permission deny/cancel/esc: a later hook that moves out of `blocked`
- user interrupt with `esc`
- user interrupt with `ctrl+c`
- session exit: `SessionEnd` reliably fires and releases the agent

if devin does not emit a reliable idle/release transition for any of those, please keep the hook session-only or state-reporting without full lifecycle authority, and let the manifest/screen fallback remain authoritative for state.

minatoaquaMK2:
> thanks, this is a strong first pass. the osc choice makes sense if the raw pty capture only showed cwd/prompt/task title and no stable lifecycle markers.
> 
> the main thing i want verified before treating `herdr:devin` as full lifecycle authority is the negative/interrupt path. since full lifecycle authority pauses screen detection, a missed hook transition can leave the pane stuck in `working` or `blocked`.
> 
> please verify the actual hook event sequence for these cases and either document it in the pr or adjust the integration:
> 
> * normal turn: `UserPromptSubmit` -> tool hooks if any -> `Stop` once the ui is idle
> * permission approve: `PermissionRequest` -> a later hook that moves back to `working`
> * permission deny/cancel/esc: a later hook that moves out of `blocked`
> * user interrupt with `esc`
> * user interrupt with `ctrl+c`
> * session exit: `SessionEnd` reliably fires and releases the agent
> 
> if devin does not emit a reliable idle/release transition for any of those, please keep the hook session-only or state-reporting without full lifecycle authority, and let the manifest/screen fallback remain authoritative for state.

Thanks for calling this out. I verified the Devin hook sequences across the normal, approval, denial, cancel, interrupt, and exit paths.

The safe cases were normal turns and permission approvals, where Devin emits a later hook that moves the pane forward. However, permission deny/esc and user interrupts with esc or ctrl+c do not reliably emit a Stop/idle transition after the UI returns to idle. SessionEnd does fire on explicit exit.

Given that, I updated the integration so Devin is no longer treated as full lifecycle authority. The hooks now refresh the native session reference for restore, while screen/OSC detection remains authoritative for state. I also updated the preview docs and change log to describe that behavior and added tests for the session-only handling and legacy lifecycle hook cleanup.

docs/next/website/src/content/docs/integrations.mdx
```diff
@@ -1,11 +1,11 @@
 ---
 title: Integrations
-description: Install Herdr integrations for Pi, OMP, Claude Code, Codex, GitHub Copilot CLI, Droid, Kimi Code CLI, OpenCode, Kilo Code CLI, Hermes Agent, Qoder CLI, and Cursor Agent CLI.
+description: Install Herdr integrations for Pi, OMP, Claude Code, Codex, GitHub Copilot CLI, Devin CLI, Droid, Kimi Code CLI, OpenCode, Kilo Code CLI, Hermes Agent, Qoder CLI, and Cursor Agent CLI.
 ---
 
 Herdr detects supported agents automatically. Official integrations can add native session identity for restore, lifecycle state reports, or both.
 
-Use integrations when you want native agent session restore, direct lifecycle reports from Pi/OMP/Copilot/OpenCode/Kilo/Hermes-style hooks or plugins, or both. See [Agents](/docs/agents/) for the full status authority model.
+Use integrations when you want native agent session restore from Claude Code/Codex/Copilot/Devin-style hooks, direct lifecycle reports from Pi/OMP/Devin/OpenCode/Kilo/Hermes-style hooks or plugins, or both. See [Agents](/docs/agents/) for the full status authority model.
```
ogulcancelik: lgtm, could you also remove devin from here since it's not lifecycle supported

ogulcancelik (changes requested):

docs/next/website/src/content/docs/integrations.mdx
minatoaquaMK2: Good catch, thanks. Updated that line so Devin is only listed for session restore, not lifecycle reporting.

ogulcancelik:
pushed a cleanup commit to keep devin session-only for now.

can you please run this PR branch with devin locally and verify the session id reporting works across start, prompt/tool use, permission request, stop, and resume?

the important rule here is: if devin hooks do not provide reliable full lifecycle authority, they should not report status states like working/blocked/idle/release. in that case hooks should only report session identity, and herdr should keep using manifest/osc/screen detection as the state authority.

minatoaquaMK2:
> pushed a cleanup commit to keep devin session-only for now.
> 
> can you please run this PR branch with devin locally and verify the session id reporting works across start, prompt/tool use, permission request, stop, and resume?
> 
> the important rule here is: if devin hooks do not provide reliable full lifecycle authority, they should not report status states like working/blocked/idle/release. in that case hooks should only report session identity, and herdr should keep using manifest/osc/screen detection as the state authority.

Thanks, pulled your cleanup and pushed a follow-up on top.

I verified the branch locally with Devin. Hooks are now session-only: the socket only saw pane.report_agent_session events, all for the same Devin session id, with no state field and no release/lifecycle reports.

One extra fix I added: startup SessionStart and UserPromptSubmit no longer fall back to devin list, since that can still point at a stale same-cwd session before the new session is established. PreToolUse / PermissionRequest / PostToolUse / Stop / resume still report the session identity correctly.

PR #685, opened 2026-06-18
feat(ui): make focused split-pane border weight configurable

asumaran:
Add a [ui] active_pane_border option ("thick" default, "plain") controlling the focused split-pane border glyph weight. The focused pane keeps the accent color in both cases; "plain" drops the heavy line to match the regular border. The setting is applied on startup and on config hot-reload.


asumaran:
Didn’t mean to create the PR here, sorry. I’ll start a discussion first.

Issue #728, opened 2026-06-21
Pane intersection corners don't use the active pane color


cullendotdev:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

When pane_gaps is false and pane_borders is on, adjacent panes share border lines correctly, but the corner where those lines meet doesn't inherit the focused pane's accent color. The intersection stays overlay0 while every line segment touching it on the focused side is accent.

<img width="2568" height="1504" alt="Image" src="https://github.com/user-attachments/assets/f5f5413d-7454-40c5-9fdb-690295fb213d" />

### Expected behavior

The active pane's color should extend through junctions, same as tmux. If the focused pane owns the corner, the corner gets accent. 

<img width="1284" height="754" alt="Image" src="https://github.com/user-attachments/assets/18aff0b3-dce0-49f8-89ec-b9b570d3fc11" />

### Reproduction

1. pane_borders true (default), pane_gaps false
2. Split the terminal into a 2×2 grid
3. Focus a pane
4. The center cross (┼) is overlay0 and the adjacent lines on the focused pane are accent

### Impact

The accent-colored lines abruptly stop at corners, so the border looks disconnected from the intersection.

### Environment

- Herdr version: master (4421c0f, unreleased)
- Update channel (stable or preview): Preview
- Operating system:  Ubuntu via WSL
- Terminal: Windows Terminal
- Shell, if relevant:
- Relevant config, if any:

cullendotdev:
`line_touches_pane` already identifies corner cells as part of a pane's border, including the explicit `shared_right && shared_bottom` corner case. The branch_cell check `line_cell_degree >= 3` then unconditionally overrides that correct answer for all junctions, contradicting the geometry model right above it.

Fix: Remove the branch_cell / line_cell_degree guard. Let line_touches_pane drive the color decision. It already handles shared-border ownership correctly. https://github.com/cullendotdev/herdr/commit/60e11b154c3bb6c22ca1f9447dd1a308f216aab2

Happy to make a PR if there's no issue with the proposed change.

ogulcancelik:
i investigated both. since terminal UIs cannot render a perfect non-overflowing junction color, the choice is either leaking the active color into neighboring borders or leaving the junction neutral. i think overlap is more common than missing junctions, so sure, please send a pr.

  /approve

PR #742, opened 2026-06-22
fix: use active pane color for border intersections

cullendotdev:
Refs #728

## Issue

When pane_gaps is false and pane_borders is on, adjacent panes share border lines correctly, but the corner where those lines meet doesn't inherit the focused pane's accent color.

## Fix
Remove the branch_cell / line_cell_degree guard. Let line_touches_pane drive the color decision.

 All tests pass:
 - 2,180 unit tests: all passed, 0 failed
 - Integration tests: all passed across all test binaries
 - clippy: clean (no warnings)
 - fmt: clean


ogulcancelik:
lgtm ty!

PR #788, opened 2026-06-24
feat: add mastracode integration

### LINKED ISSUE #337: Add support for Mastracode integration

wardpeet:
Adds mastracode integration as a TUI app.

CleanShot 2026-06-25 at 00 03 43@2x
CleanShot 2026-06-25 at 00 03 22@2x

Fixes #337

### DIFF
--- a/docs/next/api/herdr-api.schema.json
+++ b/docs/next/api/herdr-api.schema.json
@@ -1512,7 +1512,8 @@
             "kilo",
             "hermes",
             "qodercli",
-            "cursor"
+            "cursor",
+            "mastracode"
           ],
           "type": "string"
         },
@@ -5927,7 +5928,8 @@
             "kilo",
             "hermes",
             "qodercli",
-            "cursor"
+            "cursor",
+            "mastracode"
           ],
           "type": "string"
         },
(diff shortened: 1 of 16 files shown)


Roshvan:
🕉️🕉️🕉️ oh hello @wardpeet :D

ogulcancelik:
i tested the cli. my first thought was that this should be integration + manifest, but mastracode does not seem to expose good screen semantics for working state. from what i can see, the only reliable working indication is a prompt-area color change, which is not enough for a screen manifest.

so i think this needs to be a full lifecycle authority integration, similar to pi/omp. that means the hook coverage needs to be complete enough that herdr can trust it without screen fallback.

right now the pr reports idle/working/release, but it does not report blocked. we need blocked state when mastracode is waiting for user input or permission, and we need to verify that stop/abort paths fire hooks too. specifically, esc/ctrl+c while the agent is working must clear working state, otherwise herdr can get stuck showing the pane as busy.

please look at the kimi code and opencode integrations for the expected shape. opencode has easier session handling, but the important bit is the same: if mastracode cannot provide full lifecycle authority through hooks yet, then we probably need to wait until mastracode exposes the missing hook events/state.

wardpeet:
Ah sweet, thanks @ogulcancelik will fix this!

wardpeet:
Updated mastracode with all events

PR #800, opened 2026-06-25
fix: avoid pi release on internal session reloads

dmmulroy:
## Summary
- only call `pane.release_agent` for Pi `session_shutdown` with `reason === "quit"`
- keep `/reload`, `/new`, `/resume`, and `/fork` from clearing hook authority for a live pane
- refresh and report the Pi session ref on `agent_start` before publishing `working`
- bump the Pi integration asset to v4 and add regression checks

## Tests
- `just lint`
- `cargo test --locked -- --test-threads=1`

## Note
Implemented by Pi using GPT 5.5 on high, directly driven and fully reviewed by Dillon (dillon@cloudflare).


ogulcancelik:
lgtm ty! also patched omp in the same pr.

Issue #765, opened 2026-06-23
Wrong status for opencode agents


VitorLostada:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

We have 2 main issues with opencode agents.

When we run an opencode subagent and it finishes, Herdr sends a notification saying the task has been completed. That is not true, because only one subagent finished the job.

The last issue is that, when I run opencode, it sometimes keeps loading forever. 

### Expected behavior

The agent should return done only when OpenCode has finished the work.

The agent should identify the right time to stop, instead of getting stuck in an infinite loop.

### Reproduction

just ask opencode to run a sub-agent to do something, like a review. when it finishes, we will get a done message while the agent is still running

I didn't find the replicable model, but it usually happens quite often when you're not focused on the screen. I'm not sure if this is the case or if it's just something that happens inside opencode and creates this issue

### Impact

It’s really bad to use here, because I’m using it to get this status from my agents, and I basically can’t trust it. I need to manually join and recheck whether the session has finished

### Environment

- Herdr version: v0.7.0
- Update channel (stable or preview): stable
- Operating system: macoc
- Terminal: ghostty
- Shell, if relevant: zsh
- Relevant config, if any:

paperbenni:
I can reproduce the subagent issue. For me, if opencode spawns a subagent, herdr reports it as done

ogulcancelik:
did you install the integration ?

oh reprod, will fix it soon. meantime you can uninstall the integration that should work but you would lose the agent restore :/

VitorLostada:
Don't worry, I will wait for the fix. By the way, I'm falling in love with Herdr. Nice job!

VitorLostada:
After this new release, the sub-agents are fixed, but it still gets stuck in the working status forever (sometimes).

I still don’t know how to replicate it, but I’m just letting you know, @ogulcancelik 

I’ll create a new issue as soon as I figure out how to replicate it. I’m using opencode.

alecuba16:
Same I'm on 0.7.1 and it reports as idle while its working without any subagent.

PR #838, opened 2026-06-27
fix: opencode — report subagent permission prompts as blocked + handle session.status object form

soar:
## Problem

Two issues in the opencode integration plugin (`herdr-agent-state.js`) leave the pane's agent state wrong:

1. **Subagent permission prompts aren't shown as `blocked`.** When a subagent (opencode `task` tool) requests permission — e.g. a `webfetch` that needs approval — the pane stays **working** instead of **blocked**, even though the agent is waiting on the user. Root-agent prompts report correctly.
2. **`session.status` events are ignored.** opencode emits `session.status` as an object `{ type: "idle" | "busy" | "retry" }`, but `stateFromSessionStatus` only handled a bare string — so every `session.status` was a no-op, and a pane could stay stuck on **working** after an idle/resumed session.

### Reproduction (1)
1. `herdr integration install opencode`
2. In opencode, have a subagent run a tool that triggers a permission prompt (e.g. ask the `general` subagent to `webfetch` a URL, with permissions set to ask).
3. The TUI shows the prompt, but the pane / Agents panel stays `working`.

## Cause

**(1)** #765 scoped the hook to the main agent by dropping all events from child (subagent) sessions so their lifecycle churn can't clobber the pane's root state:
```js
if (sessionID && childSessions.has(sessionID)) {
  return;
}
```
That early `return` also drops a subagent's `permission.asked` / `question.asked` — exactly when it's blocked on the user. The root session is unaffected (no `parentID`).

**(2)** `stateFromSessionStatus(status)` began with `if (typeof status !== "string") return undefined`, but `session.status` carries an object, so it always returned `undefined`.

## Fix

**(1)** Before the drop, surface a subagent's blocking prompts and clear them when answered, while still dropping all other child events:
- `permission.asked` / `question.asked` → `blocked`
- `permission.replied` / `question.replied` / `question.rejected` → `working`

Reported without an `agent_session_id`, so the pane keeps the root session reference instead of adopting the subagent session.

**(2)** Read `status?.type` (handling both the object and the legacy bare string) and map `retry` → `working`.

Bumps the opencode integration version 7 → 8 (asset marker + `OPENCODE_INTEGRATION_VERSION`).

## Testing

Verified live in opencode:
- A subagent `webfetch` permission prompt flips the pane to `blocked` and back to `working` once answered; root-agent prompts behave exactly as before.
- `session.status` (object form, incl. `retry`) now keeps pane state in sync; resumed/idle sessions no longer stick on `working`.

No `src/integration` test hardcodes the opencode version, and `install_opencode_writes_plugin_to_plugins_dir` compares the installed plugin against the embedded asset.

Follow-up to #765.


ogulcancelik:
@soar hey thanks looks good but why didn't you also cover sub agent's permission asked case ? 

also one critical issue is, is there any automated permission system for sub agents in opencode? i.e. on codex there is permission hook, but main agent can review those requests and can auto approve, so as user you never see that permission request but if you add that hook that would be false positive. 

would be great if you could verify that too.

soar:
Thanks for the review! Answers to your two questions:

**1. Subagent `permission.asked` — it is covered.** That's the `permission.asked` / `question.asked` → `blocked` mapping in the child-session guard (with `permission.replied` / `question.replied` / `question.rejected` → `working` to clear it). Root prompts already worked; the fix just stops the early `return` from swallowing the subagent ones, and reports without an `agent_session_id` so the pane keeps the root session reference.

**2. Auto-approval / false positives.** I checked opencode core (`packages/opencode/src/permission/index.ts`, `Permission.ask`):
- `permission.asked` is published **only** when a requested pattern resolves to `action: "ask"`. `allow` / `deny` short-circuit **before** any event is published, and once published the call blocks on a Deferred until a human `reply()`.
- So an auto-allowed subagent permission (config `allow`, or a prior `always`) never emits the event — no false positive. Subagents go through the same resolver (with a stricter derived ruleset via `deriveSubagentSessionPermission`).
- opencode core has **no** main-agent auto-approver like codex's hook. The only auto-respond is client-side (the desktop app's `shouldAutoRespond`), where `permission.asked` is immediately followed by `permission.replied` — which this PR already maps back to `working`. So the worst case is a brief, accurate flip, never a stuck `blocked`.

I also pushed a second, related fix to this PR (description updated): opencode emits `session.status` as an object `{ type }`, but `stateFromSessionStatus` only handled a bare string, so it was a no-op — panes could stick on `working` after an idle/resumed session. Now handles both forms (+ `retry`).

ogulcancelik:
@soar okay thank thanks for clarification. please update branch (i can't ` maintainerCanModify: false` ) so we can run ci

soar:
@ogulcancelik Updated

PR #852, opened 2026-06-28
feat(keybinding): add pageup/pagedown support

sohanscript:
docs: update keybinding docs for pageup/pagedown

BREAKING CHANGE: added pageup/pagedown support to keybinding parser


sohanemon:
Thanks for the feedback. I wasn't aware of the approval workflow before opening the PR. I have already implemented this feature. Would you like me to open a Discussion first and continue from there? @ogulcancelik

Issue #895, opened 2026-06-30
Undercurl renders as plain underline in Herdr


snirt:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

In Herdr, ANSI undercurl renders as a plain flat underline.

### Expected behavior

ANSI undercurl renders as curly/wavy underline, matching Ghostty.

### Reproduction

Run in a Herdr pane:

 printf '\033[4:3mcurly?\033[4:0m\n'

### Impact

Neovim diagnostic undercurls appear as flat underlines in Herdr.

### Environment

- Herdr version:
- Update channel (stable or preview): Herdr 0.7.1 stable
- Operating system: macOS
- Terminal: Ghostty
- Shell, if relevant: Nushell
- Relevant config, if any:

snirt:
I tested current `master` at `b7015f17` locally.

The undercurl shape is fixed, but the fix seems incomplete for colored diagnostic underlines: the underline renders curly, but the underline color is still lost/incorrect in the Herdr client render path.

Repro:

```sh
printf '\033[4:3;58:2::255:255:0myellow curly underline\033[0m\n'
printf '\033[4:3;58:2::255:0:0mred curly underline\033[0m\n'
```

Expected: curly underline with yellow/red underline color.

Observed: curly underline shape works, but the underline color does not survive correctly.

I think the current fix preserves underline style by encoding it into `modifier`, but underline color is still only present in the local Ratatui/terminal style and is not carried through the semantic frame / ANSI blit protocol path. My side branch adds explicit underline color transport and made the Neovim diagnostic underline color work locally:

https://github.com/snirt/herdr/tree/issue/895-undercurl-sidebranch

snirt:
i built from master and saw that now there is white curly underline but still missing colors (red/yellow etc)

MomePP:
Reproduced independently on a different setup, same symptom as @snirt's comment above — curly shape is fixed, underline color is not.

## Environment
- herdr 0.7.3 (Homebrew release build, not `master`)
- Ghostty 1.3.2-main-+160c3c69e (tip channel), macOS 27 (Darwin 27.0.0)
- No tmux/zellij in the stack — confirmed `$TMUX`/`$ZELLIJ` unset, herdr is the only multiplexer layer involved
- `TERM=xterm-256color`

## Repro (isolated further)

[...]

PR #900, opened 2026-06-30
fix: preserve terminal underline styles

snirt:
refs #895
Before:
Screenshot 2026-06-30 at 14 38 45

After:
Screenshot 2026-06-30 at 14 39 07


snirt:
That closed but not fully fixed:
curly line added indeed but still colors are missing

Issue #947, opened 2026-07-02
kitty graphics stream freezes on partial frame updates until the window is resized


DevSrSouza:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

Streaming kitty graphics into a pane (retransmitting the same image id with new pixels a few times per second) freezes after the first frame when `kitty_graphics = true`. The program keeps running and input still reaches it, but the pane shows a stale image. Resizing the window repaints it once with the latest frame, then it freezes again.

### Expected behavior

Every retransmission of the image id replaces the visible frame, like it does when the same program runs directly in Ghostty or kitty.

### Reproduction

Enable `kitty_graphics = true` in `~/.config/herdr/config.toml`, then run this inside a herdr pane:

```python
import base64, sys, time

W = H = 200
rgba = bytearray([200, 30, 30, 255]) * (W * H)

def emit(data):
    p = base64.standard_b64encode(bytes(data))
    chunks = [p[i:i+4096] for i in range(0, len(p), 4096)]
    out = sys.stdout.buffer
    for i, c in enumerate(chunks):
        ctrl = f"a=t,f=32,t=d,s={W},v={H},i=7,q=2,m={int(i < len(chunks)-1)}" \
            if i == 0 else f"m={int(i < len(chunks)-1)}"
        out.write(b"\x1b_G" + ctrl.encode() + b";" + c + b"\x1b\\")
    out.write(b"\x1b[1;1H\x1b_Ga=p,i=7,p=1,c=20,r=10,C=1,q=2\x1b\\")
    out.flush()

emit(rgba)
for n in range(120):
    time.sleep(0.5)
    rgba[50_000] = (rgba[50_000] + 40) % 256
    emit(rgba)
```

Why it happens: the image fingerprint in `src/ghostty/mod.rs` hashes only the first, middle, and last 4096 bytes of pixel data, so same-size frames whose changes land outside those windows collide and are never re-encoded (the byte offset above is chosen to land outside them). Separately, pty writes containing kitty graphics never request an immediate render, relying on a droppable settle notify.

I have a fix for these with regression tests on my fork (`DevSrSouza/herdr`, branch `fix/kitty-graphics-streaming`, `just check` green). PR #946 was my first attempt, auto-closed for skipping this issue step. Happy to reopen it if you approve the path.

### Impact

[...]

PR #948, opened 2026-07-02
fix: kitty graphics streams freeze on partial updates and leak host images

DevSrSouza:
A program that streams kitty graphics inside a pane (retransmitting one image id with fresh pixels a few times per second) freezes after a short while with `kitty_graphics = true`. Input keeps reaching the program, but the pane shows a stale frame until the window is resized. Resizing resets the graphics surface, which is why it unsticks.

Three bugs stack up:

1. `src/pane/terminal.rs`: a pty write with a kitty graphics sequence never requests an immediate render. It relies on the 20 ms settle render, whose notify only fires if the render dirty flag was not already set. Under a continuous stream the notify can be dropped and nothing repaints.
2. `src/ghostty/mod.rs`: the image fingerprint that decides whether an image changed hashes only the first, middle, and last 4 KiB of pixel data. Same-size frames whose changes land outside those windows collide with the previous fingerprint, so the encoder emits nothing and the frame is skipped as identical.
3. `src/kitty_graphics.rs`: host image ids derive from content, so every changed frame allocates a new host image. The replaced one is never deleted. A stream leaks one full-size image into the host terminal per frame.

To reproduce (2): transmit an uncompressed image larger than 12 KiB, then retransmit it under the same id with one byte changed at an offset outside the first/middle/last 4 KiB. The attached client never sees the update.

## Fix

One commit per bug:

- Request a render on kitty graphics writes, like any other pty output. Bursts still coalesce through the dirty flag and the minimum render interval, and the settle render stays as a backstop.
- Hash the full image payload, cached per image id and recomputed only when the image's transmit time changes. libghostty-vt already refreshes the transmit time on every transmission; vendor patch 0002 exposes it through a new `GHOSTTY_KITTY_IMAGE_DATA_TRANSMIT_TIME_NS` accessor, with the removal condition listed in the patch ledger. A static image costs one map lookup per render, a streaming one costs one hash per transmitted frame. Content identity is unchanged, so identical images transmitted under different ids still share one host upload.
- Track which host image backs each pane image and delete the replaced host image once no source references it.

Each commit carries a regression test that fails without its fix. `just check` passes (2422 tests).

refs #947

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ogulcancelik (changes requested):
the overall direction looks good, including the small libghostty-vt accessor patch.

requesting changes for two cleanup items before merge:

please update `vendor/libghostty-vt.patches.md` so the new patch references `#947` instead of “to be filed”.

please also check ghostty upstream for an existing issue/pr/discussion about exposing an image transmit serial/timestamp through the libghostty-vt c api. if one exists, link it in the patch ledger. if not, please open a short upstream discussion or issue so the vendor patch has a clear removal path.

one code issue: `HostGraphicsCache.sources` needs pruning when pane/source entries disappear. as written, a stale `(pane_id, image_id)` entry can keep an old host image id “referenced” through `sources.values().any(...)`, which can preserve the exact leak the commit is trying to fix after panes close or placements disappear. prune sources against the current visible source set during `encode_graphics_update`, or otherwise remove stale entries on pane/image lifecycle.

DevSrSouza:
Thanks for the review! All three addressed:

- the ledger now references #947
- I checked ghostty upstream: the inspection C API landed in ghostty-org/ghostty#12145 and there is no transmit serial/timestamp accessor, nor any existing request for one, so I opened https://github.com/ghostty-org/ghostty/discussions/13177 and linked it in the ledger as the removal path
- `sources` is pruned against the visible source set at the start of `encode_graphics_update`, with a regression test (`stale_source_entry_does_not_block_superseded_image_delete`) for the scenario you described: a vanished source's stale entry no longer keeps the old host image referenced

I kept your master merge and pushed the changes as two commits on top. `just check` is green on the merged branch (2451 tests).

ogulcancelik:
lgtm, ty!

PR #984, opened 2026-07-03
fix: preserve pi and omp working status across reloads

dmmulroy:
## Summary

- restore Pi and Oh My Pi active state from the replacement extension context after reload
- keep active runs reported as `working` when no second `agent_start` event is emitted
- add Bun regression tests that observe the real Unix socket requests

## Tests

- `cargo fmt --check`
- `cargo nextest run --locked` (2442 passed)
- `just test-one pi_extension`
- `just test-one omp_extension`
- `just integration-assets-test`
- `just plugin-marketplace-test`

`just ci` reaches an existing `clippy::collapsible_else_if` error at `src/app/mod.rs:1564`, which this PR does not touch.


ogulcancelik:
ty

Issue #830, opened 2026-06-27
MacOS nix build fails


arunoruto:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

I tried using herdr on my MacOS system via nix, but I kept getting the same build error as on hydra: https://hydra.nixos.org/build/332986051/nixlog/1

```log
Caused by:
  process didn't exit successfully: `/nix/var/nix/builds/nix-23824-685423193/source/target/release/build/herdr-5b5e657ee3b3060e/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=vendor/libghostty-vt.vendor.json
  cargo:rerun-if-changed=vendor/libghostty-vt/build.zig
  cargo:rerun-if-changed=vendor/libghostty-vt/build.zig.zon
  cargo:rerun-if-changed=vendor/libghostty-vt/include
  cargo:rerun-if-changed=vendor/libghostty-vt/pkg
  cargo:rerun-if-changed=vendor/libghostty-vt/src
  cargo:rerun-if-changed=vendor/libghostty-vt/VERSION
  cargo:rerun-if-env-changed=LIBGHOSTTY_VT_OPTIMIZE
  cargo:rerun-if-env-changed=LIBGHOSTTY_VT_SIMD
  cargo:rerun-if-env-changed=LIBGHOSTTY_VT_ZIG_SYSTEM_DIR
  cargo:rerun-if-env-changed=HERDR_BUILD_CHANNEL
  cargo:rerun-if-env-changed=HERDR_BUILD_ID
  cargo:rerun-if-env-changed=HERDR_BUILD_COMMIT
  cargo:rerun-if-env-changed=ZIG

  --- stderr
  error: DarwinSdkNotFound
  /nix/store/rw37ij7sv3h2iic0vlk0y4078zdryjcb-zig-0.15.2/lib/zig/std/zig/LibCInstallation.zig:172:13: 0x1045a7fdf in findNative (build)
              return error.DarwinSdkNotFound;
              ^
  /nix/var/nix/builds/nix-23824-685423193/source/vendor/libghostty-vt/src/build/SharedDeps.zig:146:26: 0x1045af757 in add (build)
              const libc = try std.zig.LibCInstallation.findNative(.{
                           ^
[...]

PR #995, opened 2026-07-03
refactor(nix): replace apple-sdk wrappers with xcbuild

arunoruto:
Addresses changes introduced in the package upstreamed to nixpkgs (#830 and https://github.com/NixOS/nixpkgs/pull/536015).
The custom `apple-sdk` wrappers are not needed and can be replaced with `xcbuild`, making the package much more simple!

I tested it using `nix build .#herdr` and it worked without any problems!

Looking at the upstream package now (https://github.com/NixOS/nixpkgs/blob/a60021c1598a6551039dd3cd19607ce28e7e69f6/pkgs/by-name/he/herdr/package.nix), I am also wondering how much of the code is needed in the flake. I could also see how much can be stripped, if needed. But maybe someone with more experience can tell if I am right or I am getting something wrong :)


ogulcancelik:
ty @arunoruto since you verified it works, lgtm!

PR #1005, opened 2026-07-04
Add PageUp/PageDown keybinding support and update documentation

sohanemon:


i3130002:
This would be a great feature, at least having the pageup/pagedown as a key binding that we can assign ourselves would be amazing. 

Example key binds that works well for me:
```
[keys]
next_agent = "ctrl+down"
previous_agent = "ctrl+up"
next_tab = "ctrl+right"
previous_tab = "ctrl+left"
next_workspace = "ctrl+pagedown"
previous_workspace = "ctrl+pageup"
```


Also it might be good to capture prevent passing the rightclick and key bundings to vscode if possible.

Issue #774, opened 2026-06-24
macOS:  switch_ascii_input_source_in_prefix  switches the wrong input source


ppggff:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

With  switch_ascii_input_source_in_prefix = true  on macOS, prefix mode can leave the terminal on the wrong keyboard input source. In my setup it repeatedly switched the pane to a CJK IME (WeType pinyin) after a prefix command (e.g.  prefix +digit to switch tabs), even though the pane was on  com.apple.keylayout.ABC  the whole time.

Root cause: the prefix save/restore runs in the herdr server (a background daemon that never runs a CFRunLoop). `TISCopyCurrentKeyboardInputSource()` returns a *per-process cached* value, refreshed only when the process receives `kTISNotifySelectedKeyboardInputSourceChanged` — a CF distributed notification that is delivered solely via the **main thread's run loop**. The server pumps no run loop, so its cached value is frozen (≈ the source at startup) and the prefix-time read is stale; the restore then writes that stale source onto the foreground. (The write side works; only the read is stale.)

### Expected behavior

Prefix temporarily switches to an ASCII-capable source and restores the source the foreground *actually* has, instead of a stale cached value.

### Reproduction

Preconditions: 
* macOS, with a CJK IME plus `com.apple.keylayout.ABC`;
* switch_ascii_input_source_in_prefix = true .    

Step:
1. Launch herdr and run a prefix command while the current input source is the CJK IME.
2. In a pane, switch the input source to ABC and confirm (`macism` reports `com.apple.keylayout.ABC`).
3. Run any prefix command (e.g. `prefix`+digit to switch tabs).
4. Observed: the input source flips to the CJK IME, even though you were on ABC.

(Mirror symptom: if the server's cached source happens to be ABC while you are on a CJK IME, prefix silently does nothing instead of switching to ASCII.)

### Impact

Function not work.

### Environment

- Herdr version: 0.7.0
- Update channel (stable or preview): stable
- Operating system: macOS 26.5 (Tahoe)
- Terminal: Ghostty

ogulcancelik:
would be great if you could contribute about this since ime is hard for me to repro and verify

ppggff:
OK, I will dig deeper into the root cause and propose a feasible solution.

ppggff:
Found the root cause — updated the issue above.

[...]

PR #1016, opened 2026-07-04
fix: switch prefix ASCII input source on the foreground client

ppggff:
Fixes `experimental.switch_ascii_input_source_in_prefix` switching to or restoring a wrong/stale input source when running in the default server/client topology (#774).

### Problem

The prefix ASCII switch ran in the headless server process. `TISCopyCurrentKeyboardInputSource` returns a per-process cached value refreshed only by a distributed notification delivered through the main thread's run loop, which the headless server never pumps — so the server read a stale input source and switched/restored based on it.

### Fix

Follow the existing `ClipboardWrite` pattern and move the host-local side effect to the foreground process:

- `App::sync_prefix_input_source` emits `AppEvent::PrefixInputSource { active }` as an intent instead of touching TIS.
- The headless server forwards it to the foreground client as `ServerMessage::PrefixInputSource` (new message on the unreleased protocol 15).
- The client applies the switch/restore via `RealPrefixInputSource`, pumping its main run loop once before the TIS read so the value is fresh.
- The monolithic app (`--no-session`) applies it in-process.

### Behavior change: ASCII realm (beyond the fix itself)

The switch is now keyed on `Mode::wants_ascii_input`, an explicit allowlist covering prefix and prefix-launched command/navigation modes (navigate, navigator, copy, resize, close confirms, context/global menus, keybind help). Previously only `Mode::Prefix` itself was covered. Concretely:

- multi-level prefix commands keep the ASCII source until returning to the terminal, instead of restoring on the first sub-mode hop;
- rename and other text-entry modes restore the user's IME (they need real text input);
- modes not on the allowlist default to leaving the IME alone, so a new mode can never silently force ASCII.

Known limitation (documented in code): the navigator search box also stays on ASCII because the predicate is mode-level.

### Event-drain race (input-source part handled here)

While testing we found that `App::handle_api_request` begins with an exhaustive internal-event drain and is reachable synchronously from key handling, so a queued host-effect event can be consumed inside the server process before the forwarding drain sees it. This PR guards the input-source event with `App.local_input_source_switch` (disabled by the server, same pattern as `local_terminal_notifications`): a swallowed intent degrades to a skipped switch and recovers on the next prefix cycle. The general mechanism also affects `ClipboardWrite` (pre-existing); I will file that as a separate issue as discussed in #774.

### Protocol note

`ServerMessage::PrefixInputSource` is added to protocol 15, which is unreleased on stable (v0.7.1 = 14) but has shipped in `preview-2026-06-30`. A post-merge preview server paired with a client from that preview passes the version handshake and then fails to decode the new message (clean disconnect, no panic). If that matters for the preview channel, bumping to 16 turns it into a clean "please upgrade" error — happy to do that if preferred.

### macOS caveat (for awareness)

Independent of the stale-read fix, a bare `TISSelectInputSource` can leave the active input context stale around some CJK IMEs: the system reports the new source, but the IME's own state lags until focus leaves and returns (observed with WeType: after pinyin -> ABC the system reports ABC, yet WeType's Shift hotkey keeps toggling until a focus change). Tools in this space work around it by briefly creating a tiny key window to force an input-context refresh after the switch — macism's "TemporaryWindow" (https://github.com/laishulu/macism, applied for CJKV targets) and input-method-keeper's `refreshInputContext()` (a temporary `NSWindow` made key, closed after ~150 ms). This PR does not add that; if reports show up around specific IMEs, a follow-up could add an input-context refresh after switch/restore.

### Testing

- `just test`: 2477 nextest tests + maintenance script tests pass; `cargo fmt --check` and `cargo clippy -D warnings` clean (pre-commit).
- New tests: `Mode` classification for all 20 variants, realm persistence across multi-level prefix commands, rename restore, monolithic apply path, headless foreground-only forwarding with background negative assertion, wire roundtrip, and a regression test that the headless App never applies the switch in-process.
- Manually validated on macOS with a CJK input source: prefix chords, tab/workspace switching, rename dialogs.

refs #774

### DIFF
--- a/docs/next/CHANGELOG.md
+++ b/docs/next/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Added `ui.hide_tab_bar_when_single_tab` to hide the tab row when a workspace has one tab. (#448)
 
 ### Changed
+- Bumped the client/server protocol version to 16 for foreground-client prefix input-source switching.
 - Bumped the client/server protocol version to 15 for socket API placement mutation event and response compatibility.
 
 ### Fixed
--- a/docs/next/api/herdr-api.schema.json
+++ b/docs/next/api/herdr-api.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "protocol": 15,
+  "protocol": 16,
   "schema_version": 1,
   "schemas": {
     "error_response": {
--- a/docs/next/website/src/content/docs/configuration.mdx
+++ b/docs/next/website/src/content/docs/configuration.mdx
@@ -500,14 +500,14 @@ The trade-off when enabled: an extra hardware cursor is visible in the outer ter
 
 On macOS, prefix-mode commands can be hard to use while a non-ASCII input source is active because prefix commands are still interpreted through the host input source.
 
-Set `switch_ascii_input_source_in_prefix = true` to switch the host input source to the system ASCII-capable input source while prefix mode is active:
+Set `switch_ascii_input_source_in_prefix = true` to switch the host input source to the system ASCII-capable input source while prefix commands and prefix-launched navigation are active:
 
 ```toml
 [experimental]
 switch_ascii_input_source_in_prefix = false
 ```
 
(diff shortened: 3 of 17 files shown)


ogulcancelik:
@ppggff ty!

Issue #1017, opened 2026-07-04
Grok panes show `done/idle` while working/blocked on user input


TonyxSun:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

Grok Build panes (grok 0.2.82) basically always read idle/done. On longer
turns (subagents, shell commands) the sidebar flips back to done after a few
seconds even though grok is still running its spinner. Permission prompts and
grok's question dialogs also count as idle, so blocked never shows up.

### Expected behavior

`Working` while a turn runs, `blocked` when grok is sitting on a permission
prompt or question, `done` only when the turn actually ends

### Reproduction

1. Run `grok` in a herdr pane (Grok Build 0.2.82)
2. Give it something slow, like: spawn a subagent, then run the shell command 'sleep 45 && echo done'
3. Watch the sidebar or `herdr agent get <pane>`. It goes back to done after
   a few seconds while grok is still on "Waiting on subagent"
4. With permissions on (default mode), ask it to run: touch x.txt. The
   numbered Yes/No prompt comes up but the pane still says idle

### Impact

sidebar status broken

### Environment

- Herdr version: 0.7.1
- Update channel (stable or preview): stable
- Operating system: macOS
- Terminal: Ghostty
- Shell, if relevant: zsh
- Relevant config, if any: none

TonyxSun:
Screen recoriding of bug: 

https://github.com/user-attachments/assets/84c6e686-6f41-4484-84b4-e90edc37c475

Have a locally verified solution that i can open a PR for.

ogulcancelik:
/approve @TonyxSun 

go ahead with the pr please, appreciate the help. verified local manifest would be best way to go here for screen patterns

TonyxSun:
Thanks @ogulcancelik . Opened #1055.

TonyxSun:
https://github.com/user-attachments/assets/f088a8f0-74d7-4d07-a7a4-1de22e90eb22

perhaps need an update again :sad:

ogulcancelik:
@TonyxSun grok detection engine will be updated soon, but subagent detection requires some changes will happen only on next release i'm afraid.

PR #1055, opened 2026-07-05
fix: detect grok build working and blocked states from current ui chrome

TonyxSun:
## Summary

- Rewrite the Grok Build detection manifest for the current Grok Build UI (verified on 0.2.82). The previous rules matched pre-0.2.x chrome (`ctrl+c:cancel`, `ctrl+enter:interject`, spinner + `Run|Read|Search|List` verbs) that no longer exists, so panes fell back to idle during most working turns and never reported blocked (repro recording in #1017). https://x.ai/build/changelog shows that v0.2.x+ has been released since May 26, and it is unlikely that users are running a previous version. 
- Working turns now match the live status line (braille spinner + trailing `[stop]` chip) with an `Esc:cancel` footer-hint backstop for redraw gaps.
- Permission prompts and ask-user-question dialogs now report blocked, matched on the `┃`-guttered numbered radio option rows plus their footer hint lines.
- Idle matches the prompt footer hints (`Ctrl+.:shortcuts` without any cancel hint) as visible idle evidence.
- Keep the pre-0.2.x rules at lower priority for older Grok Build releases.
- Bump the manifest to `2026.07.03.1` / engine 2.

## State quality notes

- Evidence captured from live panes (`herdr agent read` bottom-buffer reads, `herdr agent explain --json`) while driving a real Grok Build session through responses, subagent waits, parallel tools, foreground shell runs, permission prompts, question dialogs, `/model` menus, and the shortcuts overlay.
- Working anchor requires the `[stop]` chip on the spinner line because the startup splash draws the Grok logo with braille characters; a bare braille prefix would false-positive there.
- Grok replaces the spinner with a `◆` marker while a dialog waits, so working and blocked evidence cannot overlap; blocked rules still outrank working for the parallel-tools case.
- Dialog option rows and status lines are removed from the transcript after a decision, so `whole_recent` matching does not hit stale scrollback.
- Rules are anchored to the footer hint lines via `bottom_non_empty_lines(2)` or to explicit line shapes, not whole-pane incidental text.
- Verified end to end against the running server via the local-override hot-reload loop: idle -> working (response/subagent/tool) -> blocked (permission and question dialog) -> working -> idle, plus no false positives from `/model`, the shortcuts overlay, or the splash screen.

Example captures:

```text
⠧ Waiting on subagent… 2.8s        13s ⇣29.7k [stop]        <- working
◆ Create empty blocked-test.txt file as re… [↓][stop]
┃  2 (○) Yes, proceed
┃  3 (○) No, reject (type to add feedback)
1/3:select  │  Ctrl+o:yolo  │  Ctrl+c:cancel                <- blocked
Shift+Tab:mode  │  Ctrl+.:shortcuts                         <- idle
```

## Testing

- `just check` (fmt, clippy, full nextest, windows-lint, bun and maintenance script suites)
- 13 new manifest rule tests in `src/detect/manifest/tests.rs` covering working/blocked/idle forms, the splash-logo false-positive regression, dialog-outranks-spinner priority, and the legacy-UI compatibility rules
- Manually verified against the running server through the local-override reload loop (all states above, plus the original repro from #1017 now tracking correctly)

## Notes for review

- `website/agent-detection/grok.toml` is intentionally not synced here, following #622; publishing the updated manifest to live installs via the remote fetch is a release/deploy decision (`release-docs-check` flags the version gap until synced).
- `docs/next/CHANGELOG.md` includes one Fixed line; happy to drop it if you prefer to prepare it during release review.

## Screenshots
Shows `working`, `idle`, `blocked`. 
image

refs #1017


ogulcancelik:
@TonyxSun lgtm ty! fixed some stuff, agent detection manifest should be updated live in 5~min. let me know if it works!

PR #1074, opened 2026-07-06
feat: add commandcode integration

mdeloughry:
## Summary
- add CommandCode agent detection and manifest distribution
- add CommandCode integration install/uninstall hooks for session/state reporting
- add resume support, docs, schema, and regression coverage

## Verification
- cargo fmt --check
- cargo test --locked commandcode
- python3 -m unittest scripts.test_agent_detection_manifest_check
- live-tested CommandCode working detection for Organizing/interrupt states in a dev Herdr session

Draft PR for review before opening upstream-ready PR.


ahmadawais:
https://github.com/herdrdev/herdr/discussions/3333#discussioncomment-18427961

Founder of [Command Code](https://commandcode.ai) here. Congrats on the new round @ogulcancelik 

As of https://commandcode.ai/changelog#v1.54.0 we have first-class herdr support. 

I belive y'all should ship official docs and 

image


Got `cmd` to write the detailed report for me: (agent response below):

Command Code (`cmd`) is one of the [most used coding agents for open models](https://x.com/MrAhmadAwais/status/2098600014444638368). 
It now ships herdr reporting built in: no install step, no config edit, nothing for herdr to
install into us and nothing to keep in sync with our releases.

We built it against your "Integrate your own agent" guide, and it has the four
properties that section attributes to Prime Agent's reporter:

- activates only inside herdr (`HERDR_ENV=1`, plus a pane-ownership check so a
  nested `cmd` spawned by a shell tool stands down instead of reporting into
  its parent's pane)
- maps agent events to `working`, `idle`, `blocked`
- preserves report ordering across sessions (strictly increasing `--seq`,
  seeded from the wall clock, because `cmd` re-execs on reload and relaunches
  on resume)
- releases authority on exit, on the quit path and on signals alike

It also reports session identity via `pane.report_agent_session`, in both
interactive and non-interactive runs.

What is missing is recognition. herdr has no agent id for `cmd`, so we are a
`custom:` source rather than a known agent, and there is no screen-detection
manifest. A `cmd` pane is therefore only known to herdr while our reporter is
running, and it is not running in these cases:

- the user disabled the mod (`mods.disabled: ["herdr"]` or `CMD_HERDR=0`)
- a nested `cmd` that correctly stood down because it does not own the pane
- every `cmd` released before this shipped, which is most installs today

## requested change

Two things, both small:

1. An agent id for Command Code plus a screen-detection manifest, so a `cmd`
   pane is recognized from what is on screen when no reporter is talking to
   herdr. The same fallback every other supported agent gets.

2. A `## Command Code` section in `docs/integrations.mdx`, and inclusion
   wherever supported agents are listed. Today someone running `cmd` has no way
   to learn from your docs that it works.

Explicitly NOT asking for an `herdr integration install cmd` target. We report
from inside the process, so there is nothing to install and no per-release
version skew for you to maintain. That is the part we designed to keep your
maintenance cost at zero.

If native session restore is interesting later, we already report the session
reference, so that is a follow-on rather than part of this.

## why you want this

Scale, and direct user demand.

Command Code is the most widely used coding agent for open models, running 500T+
tokens a year. We get hundreds of requests a month from users asking for herdr
support, which is why we built and shipped it rather than waiting to ask.

Those people are already your users. They run several agents side by side, which
is the workflow herdr exists for, and `cmd` is one of the agents in those panes.
Right now those panes are a hole in their sidebar whenever our reporter is not
the thing filling it.

Devs love it. 

image

Issue #681, opened 2026-06-18
copy mode blocks pane focus changes


reobin:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

After entering copy mode in one pane, I cannot move focus to another pane. Prefix-based pane navigation does not work while copy mode is active.

### Expected behavior

Copy mode should not trap pane focus. I should be able to use the normal prefix navigation to move focus to another pane, the same way I can when not in copy mode.

### Reproduction

1. Open a workspace with at least two panes.
2. Enter copy mode in one pane.
3. Press the normal prefix sequence for moving focus to another pane.

### Impact

Copy mode blocks normal pane navigation, so I have to leave copy mode before interacting with another pane.

### Environment

- Herdr version: 0.7.0
- Update channel (stable or preview): stable
- Operating system: macOS (Darwin 25.2.0)
- Terminal: Ghostty
- Shell, if relevant: zsh
- Relevant config, if any: none noted

PR #1092, opened 2026-07-06
fix: honor prefix in copy mode

reobin:
copy mode now treats the configured prefix key as a real prefix instead of handling it like a copy-mode key.

* pressing the prefix key while in copy mode opens prefix mode
* pressing escape from that prefix prompt returns to copy mode instead of exiting copy mode
* prefix + [ restarts copy mode cleanly
* pane, tab, workspace, and agent focus commands preserve copy mode when it still makes sense
* commands that leave the copy context cancel copy mode and restore scroll state

maintainer note: this gives the configured prefix key priority over copy-mode bindings. with the default prefix, `ctrl+b` now opens prefix mode while in copy mode, so copy-mode `ctrl+b` page-up only works if the user has configured a different prefix.

refs #681

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -313,6 +313,9 @@ impl AppState {
             return false;
         }
 
+        if self.copy_mode.is_some() {
+            self.clear_copy_mode_selection();
+        }
         self.switch_workspace_tab(ws_idx, tab_idx);
         if let Some(tab) = self
             .workspaces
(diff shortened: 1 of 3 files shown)


ogulcancelik:
Thanks for the PR. The core direction was right: copy mode should honor the configured prefix so prefix-mode navigation remains the escape hatch.

I landed a maintainer version in 17e2b2b878ca9b8cad60fd26baef549479c13081 rather than request a larger rewrite here. While testing the state model we found a few extra cases that needed to be handled together: removed panes/tabs clear stale copy-mode state, same-focus prefix navigation keeps active selections, scroll restore works when copy mode survives a workspace switch, and API/plugin focus paths resync copy mode too.

The landing commit references this PR and includes your co-author trailer. Thanks for catching the original conflict.

Issue #1115, opened 2026-07-07
No way to leave navigator search without losing the filter or closing the navigator


liby:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

In the session navigator, `/` focuses the search input. From there, `Esc` with a non-empty query wipes the query and state filter; `Esc` with an empty query closes the whole navigator. There is no way to leave the search input and go back to the command keys (`j/k`, `b/w/i/d/a`) while keeping the typed filter.

### Expected behavior

Some way to leave the search input without closing the navigator or losing the typed filter, so the filtered list can be navigated with the command keys.

### Reproduction

1. Open the session navigator, press `/`, type a query.
2. Press `Esc`: the query and filter are discarded.
3. Press `/` again, then `Esc` with the empty query: the navigator closes.

### Impact

Filtering then moving with `j/k` requires retyping the filter every time; `Esc` muscle memory either destroys the filter or closes the navigator.

### Environment

- Herdr version: 0.7.1
- Update channel (stable or preview): stable
- Operating system: macOS 26.5.1 & zsh
- Terminal: Ghostty
- Shell, if relevant: n/a
- Relevant config, if any: n/a

liby:
I have a [working patch](https://github.com/liby/herdr/tree/bugfix/navigator-search-escape-flow) for this: `Esc` first leaves search focus and keeps the current filter, a second `Esc` closes the navigator, and `/` resumes editing the existing query. Tests included. Happy to open a PR if the issue is accepted and the PR path is approved.

ogulcancelik:
/approve @liby 

patch looks good, thanks! you can send the pr

PR #1140, opened 2026-07-07
fix: preserve navigator search commands

liby:
Refs #1115

## Behavior

- Esc while search is focused only unfocuses the search field; the query and active state filter are kept.
- Esc outside search focus closes the navigator without clearing the query/filter first.
- `/` re-enters search focus without clearing the current query (it still clears the state filter).
- Footer now shows shortcuts for the active mode: search focus shows `enter switch`, `↑↓ move`, `ctrl+u clear`, `esc back`; command mode keeps `enter switch`, `/ search`, `b/w/i/d/a states`, `j/k/↑↓ move`, `esc close`.

## Tests

- `navigator_empty_search_escape_returns_to_commands`
- `navigator_search_escape_blurs_then_next_escape_closes`

### DIFF
--- a/src/app/input/modal.rs
+++ b/src/app/input/modal.rs
@@ -165,15 +165,7 @@ pub(crate) fn handle_navigator_key(
     if state.navigator.search_focused {
         match key.code {
             KeyCode::Esc => {
-                if state.navigator.query.is_empty() {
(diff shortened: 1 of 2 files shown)


ogulcancelik:
lgtm ty!

Issue #943, opened 2026-07-02
Pane keeps a previous agent's session forever when that agent's exit was never observed


FirstSS-Sub:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

One of my panes ran codex earlier and now runs claude. `herdr pane get` / `herdr agent get` still report the old codex `agent_session` after hours of using claude in that pane. The claude integration's SessionStart reports return ok but the stored session never changes, and `herdr pane release-agent <pane> --source herdr:codex --agent codex` also returns ok with no effect. The pane stays associated with the stale codex session.

### Expected behavior

Once a different agent has clearly taken over the pane (running in the foreground for a long time, with its official integration reporting session starts), the pane's agent_session should eventually follow the agent actually running there — or there should be some way to clear the stale one.

### Reproduction

1. Install the codex and claude integrations.
2. Set `resume_agents_on_restore = false` (only so codex is not auto-relaunched in step 5; my real pane got into this state with it enabled).
3. Run codex in a pane and wait until `herdr pane get <pane>` shows its agent_session.
4. While codex is still running, restart the herdr server (`herdr server stop`, then start herdr again), so codex's exit is never observed.
5. Run claude in that pane and use it.
6. `herdr pane get <pane>` keeps showing the codex agent_session indefinitely. The release-agent call above returns ok but changes nothing.

### Impact

My main work pane is permanently tied to a stale codex session even though I have been working in claude there for hours; with resume on restore enabled the pane would relaunch the wrong agent. I found no CLI/API way to recover short of closing the pane.

### Environment

- Herdr version: 0.7.1
- Update channel (stable or preview): stable
- Operating system: macOS 15 (Darwin 24.6.0)
- Terminal: Ghostty
- Shell, if relevant: zsh
- Relevant config, if any: claude integration v7, codex integration v6, resume_agents_on_restore = true

ogulcancelik:
congrats on finding a very good edge case 😀 fixed on master for now.

iRonin:
Cheers. This was bugging me...
#667

FirstSS-Sub:
@ogulcancelik 
Thanks for the quick fix!
[...]

PR #1189, opened 2026-07-08
fix: reanchor pi status after session replacement

dmmulroy:
## Summary

- forward Pi session lifecycle reasons with session reports
- reanchor full-lifecycle authority for Pi startup, new, resume, and fork transitions
- allow an explicitly resumed Pi session to leave the stale-session set while keeping the abandoned session stale
- bump the Pi integration to v5 and verify bundled asset versions match their runtime constants

Without this, Herdr can acknowledge Pi reports after a session replacement while retaining the previous JSONL path, leaving a visibly working Pi pane reported as idle or done.

## Tests

- `just ci`
- focused session replacement, stale resume, negative lifecycle, and v4 upgrade tests
- adversarial pre-fix reproduction and post-fix validation

refs #943

### DIFF
--- a/src/integration/assets/pi/herdr-agent-state.ts
+++ b/src/integration/assets/pi/herdr-agent-state.ts
@@ -2,7 +2,7 @@
 // managed by herdr; reinstalling or updating the integration overwrites this file.
 // add custom hooks/plugins beside this file instead of editing it.
 // HERDR_INTEGRATION_ID=pi
-// HERDR_INTEGRATION_VERSION=4
+// HERDR_INTEGRATION_VERSION=5
 // @ts-nocheck
 
 import { createConnection } from "node:net";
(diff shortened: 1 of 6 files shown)


dmmulroy:
i will conquer the pi integration and all of the edge cases one of these times

src/integration/assets/pi/herdr-agent-state.ts
```diff
@@ -336,13 +337,13 @@ export default function (pi) {
...
 
-  pi.on("session_start", (_event, ctx) => {
+  pi.on("session_start", (event, ctx) => {
     if (ctx?.hasUI !== true) {
       return;
     }
     rootSession = true;
     updateSessionRef(ctx);
-    void reportSession();
+    void reportSession(event?.reason);
```
ogulcancelik: hmm could this race with below `publishState(true);`? since we don't wait it, state update could arrive first then herdr would drop reporting the new session again. maybe we should wait this?

ogulcancelik:
> i will conquer the pi integration and all of the edge cases one of these times

pi went from one of the easiest agents to build an integration on top of to one of the most edge case heavy pretty fast :/ but I appreciate the help!

src/integration/assets/pi/herdr-agent-state.ts
dmmulroy: fixed 🫡 thanks for the review!

Issue #912, opened 2026-07-01
`new_cwd = "follow"` does not apply to new tabs on Windows


gursuj:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

New tabs always open in the workspace's initial directory. This happens with and without `new_cwd = "follow"` set explicitly. Pane splits (prefix+v, prefix+-) correctly inherit the active pane's cwd.

### Expected behavior

New tabs should follow the source pane's cwd, the same as pane splits do and as the docs describe for `new_cwd = "follow"`.

### Reproduction

1. Create a workspace from any directory, e.g. `C:\foo`
2. Navigate to a different directory in a pane, e.g. `C:\bar`
3. Press prefix+c (or run herdr tab create) — new tab opens in `C:\foo`

### Impact

Every new tab requires manual navigation back to the working directory. Having it follow cwd as per configuration would be convenient

### Environment

- Herdr version: 0.7.1-preview.2026-06-30-3459798b606d
- Update channel: preview
- Operating system: Windows 11 Pro 10.0.26200
- Terminal: Windows Terminal
- Shell: PowerShell 7 (specified via `default_shell = "C:\\Program Files\\PowerShell\\7\\pwsh.exe" under [terminal]`
- Relevant config: `new_cwd = "follow" under [terminal]`

gursuj:
Might be somewhat related? When herdr server is restarted and reconnected, restored panes, tabs start in initial directory instead of the last directory they were in. Especially an issue when using `resume_agents_on_restore`

**Steps:**
1. Start a herdr session/workspace from directory A.
2. `cd` into a different directory B, then start `claude` there (agent session gets recorded against pane B).
3. Stop the herdr server (crash, `herdr server stop`, or reboot) and relaunch `herdr` so it reattaches the saved tabs/panes and runs the configured resume command (`claude --resume <id>`).

**Current behavior:** The tab restores into directory A (the workspace's starting dir), not directory B (the pane's last actual cwd). Because `claude --resume` requires being invoked from the same cwd the session was created in, resume fails when it's launched from A instead of B.

**Expected behavior:** Restored panes should use the pane's last known cwd (B), matching what session.json actually recorded for that pane while it was live — the same expectation as the new-tab case above.

gursuj:
Thank you! I was eagerly waiting for this. Works well now.

gursuj:
[...]

PR #1260, opened 2026-07-10
fix(terminal): honor in-pane cd when new_cwd=follow

### LINKED ISSUE #1245: `new_cwd = "follow"` does not seem to apply my previous cwd to new panes

### Current behavior

New panes open at the workspace root always.

### Expected behavior

With `new_cwd = "follow"`, new panes/tabs should open in the directory I'm currently in, instead of the workspace root. https://github.com/ogulcancelik/herdr/issues/912 looks very similar to what I'm seeing, but on ver 0.7.3, I still see the buggy behavior.

### Reproduction

1. open a pane at `~/root_folder`.
2. `cd path/to/folder` (start a nested shell in a subdirectory).
3. open a new pane/tab (prefix+v or prefix+c).
4. new pane opens at `root_folder`, not `root_folder/path/to/folder`.

### Impact

I work across several folders at once and therefore I need to be able to know that my panes open in the folder that I expect without having to navigate after-the-fact. Every new pane/tab I open afterward lands in the wrong place, so I have to manually `cd` every time. The rest of the tool works great but this is enough of a deal to me that I will continue using tmux.

### Environment

- herdr 0.7.3, stable channel
- OS: NixOS (Linux)
- Terminal: Ghostty
- Shell: zsh
- Config: `[terminal] new_cwd = "follow"`

Pitchfork-and-Torch:
## Summary

Fixes #1245 (related: #912).

With `terminal.new_cwd = "follow"`, new panes/tabs often still opened at the **workspace / pane launch directory** after the user `cd`s inside the focused pane. That matches using OSC-7 / shell-leader cwd only, which frequently does **not** update for nested shells.

## Change

Prefer the **foreground process cwd** (`foreground_cwd_for_pane` / `process_cwd` of the controlling process group) when resolving follow sources, then fall back to the existing OSC/cached `cwd_for_pane` path.

Applied on:

- Focused-pane follow helper used by tab create / workspace create
- Pane split API
- Layout apply / layout split
- In-process `AppState::split_pane` (keybind path)

## Why this is correct

`cwd()` tracks reported/shell-leader directory. `foreground_cwd()` reads the live process group controlling the PTY — the path after `cd` in an interactive nested shell. Config docs already describe follow as inheriting the source pane’s current directory; this makes that true for real shell navigation.

## Test plan

- [ ] Config: `[terminal] new_cwd = "follow"`
- [ ] Open pane at workspace root → `cd path/to/subdir` → prefix+c / prefix+v → new tab/split starts in `subdir`
- [ ] With OSC-7 shells that already report cwd, behavior unchanged
- [ ] Existing unit tests for follow/cached cwd still pass (`cargo test new_terminal_cwd` / workspace_create_follows_…)

## Notes

- Foreground cwd is Unix-capable via platform `process_cwd`; on platforms where it is unavailable, behavior falls back to the previous cached path.
- No config schema change.

### DIFF
--- a/src/app/api/layouts.rs
+++ b/src/app/api/layouts.rs
@@ -341,11 +341,12 @@ impl App {
         let follow_cwd = replace_target.and_then(|(_, tab_idx)| {
             let ws = self.state.workspaces.get(ws_idx)?;
             let tab = ws.tabs.get(tab_idx)?;
-            tab.cwd_for_pane(
-                tab.layout.focused(),
-                &self.state.terminals,
-                &self.terminal_runtimes,
-            )
+            let pane_id = tab.layout.focused();
+            // Prefer live foreground cwd (nested shell after `cd`) over OSC/cache (#1245).
+            tab.foreground_cwd_for_pane(pane_id, &self.terminal_runtimes)
+                .or_else(|| {
+                    tab.cwd_for_pane(pane_id, &self.state.terminals, &self.terminal_runtimes)
+                })
         });
         self.resolve_new_terminal_cwd(
             follow_cwd.or_else(|| self.focused_pane_cwd_in_workspace(ws_idx)),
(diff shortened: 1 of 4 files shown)


Pitchfork-and-Torch:
Understood — thanks for the clear process note.

I opened this against the existing bug report #1245 (reproducible \
ew_cwd = "follow"\ after in-pane \cd\). I'll request first-time approval on that issue and won't re-open a PR until a maintainer marks \/approve @Pitchfork-and-Torch\ per CONTRIBUTING.md.

Happy to adjust or drop the branch if the approach doesn't fit herdr's design.

PR #1261, opened 2026-07-10
fix(render): emit SGR 58 underline color to the host terminal

### LINKED ISSUE #1252: Child pane SGR 58/59 underline colors do not affect visible underlines

### Current behavior

Inside a herdr pane, child processes can render underlined text, including extended underline styles such as undercurl, but SGR 58 underline colors do not affect the visible underline color. The underline uses the text foreground/default color instead.

The same commands render colored underlines when run directly in Ghostty outside herdr.

> [!NOTE]
> This is related to, but different from, #1234. That report covers OSC 12/112 cursor color. This report covers SGR 58/59 cell underline color.

### Expected behavior

When a child emits SGR 58, the visible underline should use the requested indexed or RGB color independently of the text foreground.

When the child emits SGR 59, the underline color should return to the terminal default.

### Reproduction

Run each command directly in the host terminal first, then run the same command inside a herdr pane.

Render white text with a red curly underline:

```shell
printf '\033[38;2;255;255;255;4:3;58;2;255;95;95mWHITE TEXT / RED UNDERCURL\033[0m\n'
```

Render white text with a green underline, then reset only the underline color:

```shell
printf '\033[38;2;255;255;255;4;58;2;95;255;135mGREEN UNDERLINE\033[59m DEFAULT UNDERLINE\033[0m\n'
```

Observed comparison:

- Outside herdr: the first underline is red; the second command changes from a green underline to the default underline color.
- Inside herdr: the underline style is visible, but the requested underline colors are not reflected in the visible output.

### Impact

Editors and terminal applications use independently colored underlines and undercurls for diagnostics, spellchecking, and status decoration.

Inside herdr, the underline remains visible but loses its semantic color. For example, Neovim diagnostics cannot display their configured colored undercurls even when the outer terminal supports them.

### Environment

- Herdr version: `herdr 0.7.2-preview.2026-07-07-f5354780e4ef`
- Update channel (stable or preview): preview
- Operating system: macOS 26.5.2 (25F84)
- Terminal: Ghostty (tip; build 16474)
- Shell, if relevant: fish, version 4.8.0
- Relevant config, if any: n/a

Pitchfork-and-Torch:
## Summary

Fixes #1252.

Inside a herdr pane, underline **styles** (including undercurl `4:3`) already rendered, but **SGR 58 underline colors** did not. Colored underlines always fell back to the text foreground / host default, so Neovim diagnostics and similar decorations lost semantic color. Ghostty outside herdr showed the colors correctly.

## Root cause

1. Internal VT (libghostty) + ratatui conversion already kept `underline_color` (covered by `render_preserves_underline_color`).
2. Wire `CellData` only carried `fg` / `bg` / `modifier` — **no underline color**.
3. Host ANSI blit (`build_sgr`) never emitted **SGR 58** (or 59), so even local host rendering lost the color when frames were re-emitted as escape sequences.

## Change

| Area | Change |
|------|--------|
| `CellData` | Add `underline: u32` (same packing as `fg`/`bg`), `#[serde(default)]` for older peers |
| Buffer → frame | Copy `Style.underline_color` into `CellData.underline` |
| Dirty patch / style | `cell_data_from_style` includes underline color |
| `build_sgr` | Emit `58;2;r;g;b` / `58;5;n` when set (SGR `0` already resets to default) |
| Diff equality | Treat underline color as a visual difference |

## Tests

- Existing: `render_preserves_underline_color` (buffer path)
- New: `color_to_sgr_ul_rgb_and_indexed`, `build_sgr_includes_underline_color`

## Manual verify

```sh
printf '\033[38;2;255;255;255;4:3;58;2;255;95;95mWHITE TEXT / RED UNDERCURL\033[0m\n'
printf '\033[38;2;255;255;255;4;58;2;95;255;135mGREEN UNDERLINE\033[59m DEFAULT UNDERLINE\033[0m\n'
```

Inside a herdr pane (host Ghostty/Kitty/etc. with underline color support): underline should be red / green independently of white text.

## Notes

- No protocol version bump: field is additive with serde default `0` (default underline color).
- Foreground-only underlines (no SGR 58) behave as before.

### DIFF
--- a/src/pane/terminal.rs
+++ b/src/pane/terminal.rs
@@ -2007,6 +2007,9 @@ fn cell_data_from_style(symbol: String, style: Style) -> CellData {
         },
         fg: crate::protocol::color_to_u32(style.fg.unwrap_or(Color::Reset)),
         bg: crate::protocol::color_to_u32(style.bg.unwrap_or(Color::Reset)),
+        underline: crate::protocol::color_to_u32(
+            style.underline_color.unwrap_or(Color::Reset),
+        ),
         modifier: crate::protocol::modifier_to_u16(style.add_modifier),
         skip: false,
         hyperlink: None,
--- a/src/protocol/render_ansi.rs
+++ b/src/protocol/render_ansi.rs
@@ -291,6 +291,30 @@ fn color_to_sgr_bg(val: u32) -> String {
     }
 }
 
+/// Converts a packed u32 color to an SGR 58 underline-color fragment.
+///
+/// Returns `None` for default/reset underline color (after a leading SGR `0`,
+/// underline color is already default — SGR 59 is only needed when changing
+/// away from a previous non-default color without a full reset).
+fn color_to_sgr_ul(val: u32) -> Option<String> {
+    match val >> 24 {
+        0x00 => match val & 0xFF {
+            0x00 => None, // Reset / default
+            // Map named colors onto the standard 16-color palette for 58;5;n.
+            n @ 0x01..=0x10 => Some(format!("58;5;{}", n - 1)),
+            _ => None,
+        },
+        0x01 => Some(format!("58;5;{}", val & 0xFF)),
+        0x02 => {
+            let r = (val >> 16) & 0xFF;
+            let g = (val >> 8) & 0xFF;
+            let b = val & 0xFF;
+            Some(format!("58;2;{r};{g};{b}"))
+        }
+        _ => None,
+    }
+}
+
 // ---------------------------------------------------------------------------
 // Modifier → SGR
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


Pitchfork-and-Torch:
Understood — thanks for flagging the first-time contributor gate.

This was aimed at the open bug #1252 (SGR 58/59 underline color lost on host re-emit). I'll ask for approval on that issue and wait for \/approve @Pitchfork-and-Torch\ before any further PR.

No problem if you'd rather handle it in-tree differently.

robfeldmann:
Thanks for working on this! I'm not an approved contributor (or an experienced Rust dev), but I will say this solution aligns with my own investigation of root cause and potential fix.

PR #1414, opened 2026-07-14
feat: add OSC 52 clipboard query support for neovim compatibility

str0g:
OSC 52 clipboard writes already worked through herdr, but queries (ESC ] 52 ; c ; ? BEL) were silently dropped. Neovim uses this protocol to read the host clipboard from within panes, so it was broken — paste-from-host operations failed silently.

Claude Desktop / Ollama CLI happen to work fine (they rely on writes or a different exchange path), which is why this went unnoticed until neovim usage surfaced the gap.

This commit adds bidirectional OSC 52 clipboard support:

- The OSC 52 forwarder now distinguishes write sequences from query sequences using a new ClipboardWriteOrQuery enum, tracking both flows separately through the PTY read path.

- When a child process sends a clipboard query, herdr reads the system clipboard via the platform API, base64-encodes it, and replies with an OSC 52 answer sequence (ESC ] 52 ; a ; c ; <base64> BEL).

- Both TUI client mode and headless server mode handle the new ServerMessage::ClipboardQuery wire variant.

- Wire protocol gets the new ClipboardQuery message type; selection module gains osc52_query_response_sequence() and respond_to_clipboard_query() helpers.

Implementation vibe-coded using Qwen 3.6 35B served locally through Ollama and Claude.

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -2705,6 +2705,7 @@ impl AppState {
             // or via HeadlessServer forwarding to the foreground client (server); never touch
             // AppState. Kept for AppEvent exhaustiveness.
             AppEvent::ClipboardWrite { .. } => Vec::new(),
+            AppEvent::ClipboardQuery { pane_id: _ } => Vec::new(),
             AppEvent::PrefixInputSource { .. } => Vec::new(),
             AppEvent::TerminalCwdReported { pane_id, cwd } => {
                 if !cwd.is_absolute() || !cwd.is_dir() {
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -1556,6 +1556,15 @@ async fn run_client_loop(
                     forward_clipboard(&data);
                     let _ = io::stdout().flush();
                 }
+                ServerMessage::ClipboardQuery { data } => {
+                    // Child sent an OSC 52 clipboard query; decode and write back
+                    // an OSC 52 query response (`ESC ] 52 ; a ; c ; <base64> BEL`).
+                    if let Some(bytes) = decode_clipboard_payload(&data) {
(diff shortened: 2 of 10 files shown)


str0g:
test system archlinux, neovim 0.12.

str0g:
AI slop review... if any of those comments were true copy  and passing from to neovim would not work.

benkoppe:
Please god @ogulcancelik ik no one uses editors anymore because of AI and all that but I would kill for this feature lol

ogulcancelik:
the ai review was correct. the patch detects the query, but writes its answer to herdr’s outer stdout instead of the requesting child’s pty. it also sends the invalid form `osc 52;a;c;<data>` instead of `osc 52;c;<data>`, loses the originating pane in server mode, has no response route back to that pane, and changes the wire protocol without the required version bump. the parser tests do not test this end-to-end path.

this is also not standard multiplexer behavior. neovim inside tmux normally uses tmux’s `load-buffer`/`save-buffer` provider or ordinary terminal paste. tmux’s osc 52 query support defaults to returning its own buffer; requesting the real terminal clipboard requires explicit configuration. zellij desktop still ignores osc 52 queries entirely.

the proposed behavior would let any process inside a pane silently read the host clipboard. this patch reads it directly without an opt-in, prompt, or policy, bypassing the outer terminal’s clipboard-read protections.

you can use neovim’s native `wl-copy`, `xclip`, or `pbcopy` provider when available, or paste through the terminal with `ctrl-shift-v`/`cmd-v`; osc 52 copy from neovim already works. remote users can use normal terminal paste, but direct `"+p` through an osc 52 clipboard query is currently unsupported.

for those reasons, i’m not going to reopen this pr or request a replacement in its current form. if you have a better approach that addresses routing and clipboard security, please open a discussion.

Issue #1000, opened 2026-07-04
copy mode horizontal movement stops mid-glyph on full-width (CJK) characters


m1ng112:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

In copy mode (prefix+[), moving horizontally with l/h over a full-width (CJK, e.g. Japanese) character requires two key presses to cross a single character. The cursor stops on the trailing half of the wide glyph after the first press.

### Expected behavior

One l/h press moves the cursor across a whole full-width character, the same as it does for a half-width ASCII character.

### Reproduction

  1. In a pane run: printf 'AXあYb\n'
  2. Enter copy mode with prefix+[
  3. Move the cursor onto the あ character
  4. Press l once — the cursor lands on the trailing cell of あ instead of moving to the next character Y

### Impact

Horizontal navigation and text selection over Japanese/CJK text in copy mode is off by one cell per wide character, making selection tedious and error-prone.

### Environment

- Herdr version: 0.7.1
- Update channel (stable or preview): stable
- Operating system: Mac Tahoe  version 26.5
- Terminal: Wezterm 20240203-110809-5046fc22
- Shell, if relevant: zsh
- Relevant config, if any:

blokboy:
Hello, I would be interested in submitting a fix that I believe will cover this as well as Issue #970. Since this would be my first time contributing, I'm commenting here instead of opening a separate discussion per contrib guidelines. Adding a bit more information around the changes I'd like to propose as well as [my working branch ](https://github.com/blokboy/herdr/tree/fix/copy-mode-wide-word-motions)  

Issue #1000:

Horizontal cursor movement now treats full-width characters as a single logical character instead of two independent cells. Previously, pressing l or the right arrow over a CJK character could land the copy cursor on the trailing spacer cell, so moving across あ required two key presses. The new helper snaps horizontal movement to character boundaries: moving right from the leading cell advances past the full-width glyph, and moving left from its trailing cell returns to the glyph start. This keeps cursor movement aligned with visible characters rather than raw terminal cells.

**Patch**: 

My patch changes copy-mode cursor math in copy_mode.rs (line 257)

Before this, horizontal motion was pure column arithmetic: cursor_col += 1 or cursor_col -= 1. That works for ASCII, but terminal cells are not always one character. A CJK character like あ is one character occupying two display cells. So the old code could move the cursor onto the second half of the glyph.

[...]

PR #1469, opened 2026-07-16
fix: step copy mode horizontal motion over full-width glyphs

kovalov:
Fixes copy-mode horizontal movement stopping mid-glyph on full-width (CJK) characters.

`move_copy_cursor` stepped horizontal motion by raw columns, so `h`/`l` landed on the trailing cell of a wide glyph and needed two presses to cross a single character. This steps horizontal motion by whole glyphs: it reads the visible row text and advances/retreats to the next character boundary via `row_cell_glyphs` / `next_cell_col` / `prev_cell_col`, falling back to single-cell steps when the row text is unavailable.

Reproduction from the issue (`printf 'AXあYb\n'`): cursor on the leading cell of `あ` (col 2), one `l` now moves to `Y` (col 4) instead of the trailing cell, and one `h` returns to col 2.

Tests:
- `cell_col_steps_handle_wide_glyphs_and_blanks` (unit)
- `copy_mode_horizontal_crosses_full_width_glyph` (integration)

refs #1000

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved horizontal cursor movement in copy mode.
  * Wide characters are now traversed as a single unit instead of stopping on trailing spacer cells.
  * Combining marks and blank cells are handled more accurately.
  * Cursor movement now remains within valid row boundaries.


kovalov:
@ogulcancelik Pls review.  I have resolved the conflict.

ogulcancelik:
@kovalov could you update the pr please?

PR #1648, opened 2026-07-20
fix: scope pi and omp releases to active sessions

dmmulroy:
## Summary

Fix nested Pi and OpenCode OMP runtimes incorrectly releasing or replacing the active root agent session by scoping lifecycle releases to the session that owns the pane.

A matching session-scoped release still clears immediately. This implementation does not add process-owner tracking, termination confirmation events, runtime claim mutexes, or handoff plumbing.

## What changed

- Add optional `agent_session_id` and `agent_session_path` fields to `pane.release_agent`, its CLI wrapper, and the unreleased API schema/docs.
- Validate release source, agent, and session ownership in `TerminalState` before accepting a sequence number.
- Ignore mismatched scoped releases and unscoped releases against session-bound ownership without consuming their sequence number.
- Preserve immediate release behavior for custom and other sessionless integrations.
- Include `currentSessionRef()` in Pi and OMP release requests.
- Include MastraCode's known `session_id` in release requests and bump its integration migration version from the latest released value.
- Validate session-start semantics before accepting sequence numbers, so rejected nested starts cannot block later root updates.
- Stop treating OMP `startup` as authorization to replace an existing session.
- Preserve an authoritative root when a different nested foreground agent is detected; a return to the pane shell remains exit evidence.

## Test coverage

Adds coverage for:

- Mismatched nested Pi and OMP releases.
- Matching root-session releases.
- Unscoped releases against hook-authority and persisted session owners.
- Custom/sessionless immediate releases.
- Rejected release and session-start sequence handling.
- Nested OMP startup behavior.
- Nested foreground agents under full-lifecycle authority.
- Pi, OMP, and MastraCode release payload session fields.
- API schema round-tripping and CLI surface coverage.

## Validation

- `just lint`
- `just integration-assets-test`
- Focused Rust test groups for terminal state, app actions, API schema, pane detection, integrations, and CLI specification.
- The full Cargo nextest phase completed with 2,854 passing tests. The broader maintenance-script stage was not rerun because it hangs in this environment.

### DIFF
--- a/docs/next/api/herdr-api.schema.json
+++ b/docs/next/api/herdr-api.schema.json
@@ -2691,6 +2691,18 @@
             "agent": {
               "type": "string"
             },
+            "agent_session_id": {
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "agent_session_path": {
+              "type": [
+                "string",
+                "null"
+              ]
+            },
             "pane_id": {
               "type": "string"
             },
--- a/docs/next/website/src/content/docs/socket-api.mdx
+++ b/docs/next/website/src/content/docs/socket-api.mdx
@@ -689,6 +689,23 @@ Session-only official integrations report native session references with `pane.r
 
 If no native session reference is stored, the field is omitted.
 
+Lifecycle integrations release agent authority with `pane.release_agent`. Integrations that know their current session should include the same `agent_session_id` or `agent_session_path` used in their state reports:
+
+```json
+{
+  "id": "req_3",
+  "method": "pane.release_agent",
+  "params": {
+    "pane_id": "w1:p1",
+    "source": "herdr:pi",
+    "agent": "pi",
+    "agent_session_path": "/path/to/pi-session.jsonl"
+  }
+}
+```
+
+When the current owner is session-bound, Herdr accepts the release only when its source, agent, and session reference all match. Mismatched or unscoped releases are ignored without consuming their sequence number. Sessionless integrations retain the existing immediate release behavior.
+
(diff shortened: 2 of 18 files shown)


ogulcancelik:
@dmmulroy what's the problem you are having with pi integration for statuses? do you have an exact repro case?

ultra2207:
omp shows tick when its still running is the issue @ogulcancelik

dmmulroy:
@ogulcancelik  I don’t have a one-command UI repro yet, but I have a deterministic server-level repro.

 The problem occurs when a Pi session starts a second Pi session in the same pane:

 1. The parent Pi reports that it is working.
 2. The nested Pi starts and inherits the same Herdr pane ID.
 3. The nested Pi exits and reports that Pi was released.
 4. The parent Pi continues working.

Herdr applies the nested Pi’s exit to the parent. It then shows the pane as idle and ignores later updates from the parent.

One of the regression tests should reproduce that sequence.

ogulcancelik:
@dmmulroy which extension or workflow is launching the nested pi process? i use https://github.com/ogulcancelik/pi-extensions/tree/main/packages/pi-codex-subagents, but its children run in rpc mode with extensions disabled, so they do not load the herdr integration and i cannot reproduce this. i still need an exact deterministic repro case to properly handle this

ogulcancelik:
@dmmulroy i think i found the problem which required a small refactor on agent detection ownership. if you are willing, you can test from master with new herdr version + new extension (tho old one should be fine with new codebase too) [e608a75](https://github.com/ogulcancelik/herdr/commit/e608a75159fc5cca70485d83ffb8ee681c9da385)

or you need to wait next week to test :/

Issue #1800, opened 2026-07-23
Contribution proposal: Grok CLI integration (session restore) + detection manifest fix


ogulcancelik:
### Discussed in https://github.com/ogulcancelik/herdr/discussions/991

<div type='discussions-op-text'>

<sup>Originally posted by **carlesso** July  4, 2026</sup>
### idea / problem

The current detection for grok cli is not working with the latest version, and there is no integration.

### requested change

Fix the detection and add the integration for session management.

### why you want this

With these fixes, grok cli works pretty darn good inside herdr!
Also I already got my grok to fix the manifest and create the integration, and I've been rocking it for more than a week, and has been working very well! Love herdr!</div>

PR #1807, opened 2026-07-24
feat: add grok cli integration for native session restore

carlesso:
Implements the integration half of #1800 (approved contribution). Adds `herdr integration install grok` following the cursor pattern: a self-contained `hooks/herdr.json` plus hook script under `~/.grok/hooks/` (grok merges all hook files, so nothing else is ever touched). The SessionStart hook reports the grok session id over the socket API, and native restore resumes panes with `grok --resume <id>`. Session-identity tier only — lifecycle state stays screen-detected.

**Note on scope**: #1800 also covered the detection-manifest fix, but `master` has since shipped its own grok detection rework (#1055 and the OSC-signal rules), which supersedes what I had — so this PR is integration + restore only.

**Testing**: full `just check` equivalent on macOS (nextest suite, maintenance script tests, integration asset tests, windows-target clippy), plus daily real-world use since July 8 across ~10 concurrent grok panes, including server restarts with native session restore.

**Docs**: `docs/next/` integrations, session-state, cli-reference pages and changelog; API schema artifact regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added Grok CLI integration support, including `herdr integration install grok` and `herdr integration uninstall grok`.
  * Grok CLI now reports session identifiers to enable native session restoration, including resuming panes with `grok --resume <id>`.
  * Integration health now marks Grok as **Outdated** when its hook configuration is missing or invalid (reinstall repairs it).
* **Documentation**
  * Updated supported agents, integrations/restore guidance, CLI reference examples, and API schema to include Grok CLI.

### DIFF
--- a/docs/next/CHANGELOG.md
+++ b/docs/next/CHANGELOG.md
@@ -125,6 +125,7 @@
 - Added `herdr terminal session control` for bridge processes that need live ANSI frames plus input, resize, scroll, release, and takeover authority.
 - Added `ui.hide_tab_bar_when_single_tab` to hide the tab row when a workspace has one tab. (#448)
 - Added Japanese and Simplified Chinese website docs.
+- Added `herdr integration install grok` for Grok CLI (Grok Build) hooks that report session ids through Herdr's socket API. Grok state stays screen-detected. When native agent session restore is enabled, Herdr can resume Grok panes with `grok --resume <id>`.
 
 ### Changed
 - The mobile switcher now starts from an agents-first summary and renders worktrees as a tree, making narrow terminals easier to scan.
--- a/docs/next/api/herdr-api.schema.json
+++ b/docs/next/api/herdr-api.schema.json
@@ -2085,7 +2085,8 @@
             "hermes",
             "qodercli",
             "cursor",
-            "mastracode"
+            "mastracode",
+            "grok"
           ],
           "type": "string"
         },
@@ -6997,7 +6998,8 @@
             "hermes",
             "qodercli",
             "cursor",
-            "mastracode"
+            "mastracode",
+            "grok"
           ],
           "type": "string"
         },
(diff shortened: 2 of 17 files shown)


ogulcancelik (changes requested):
thanks for the contribution! the session-only scope looks right: report the grok session id for native restore, while keeping screen detection as the lifecycle authority.

before merging, could you please address the two existing inline findings around honoring `GROK_HOME` and validating both the hook script and `hooks/herdr.json` in integration status?

there’s no need to add lifecycle state hooks here. grok’s current hooks don’t reliably cover interrupts and interaction resolution, so keeping this integration session-only is the safer approach.

once those are addressed, this should be ready for another pass.

carlesso:
Thanks for the review! Both findings addressed in 6bda070:

- `grok_dir()` now honors `GROK_HOME` (grok's documented config home) after the `GROK_CONFIG_DIR` test seam, with tests covering both precedences.
- Integration status now validates `hooks/herdr.json` next to the hook script — missing, corrupt, or no-longer-referencing configs report `outdated` so a reinstall repairs the pair, instead of masking a hook grok never invokes. Added degradation tests for all three cases.

Agreed on keeping this session-only — no lifecycle hooks here.

carlesso:
Good catch @greptile-apps — the substring check was indeed too loose. Status validation now compares the SessionStart entry against the exact command the installer writes (shared builder), and the degradation tests cover both nonfunctional shapes (script name mentioned but not invoked; invoked without the `session` action). Fixed in the latest commit.

carlesso:
Also addressed the follow-up finding: the validator now requires `type = "command"` in addition to the exact generated command string, with a test for the mismatched-type shape. At this point the status check accepts exactly what the installer writes and nothing else.

ogulcancelik:
ty

ecarlesso-xAI:
Thank you!!!

PR #1839, opened 2026-07-24
feat: add Antigravity session restore support

adamkrellenstein:
- Registers `"herdr:agy" / "agy"` as an official agent source.
- Formulates native session resume plan: `agy --conversation <session_id>`.
- Adds unit test coverage for `agy` resume planning and state reservation.

(Cf. https://github.com/ogulcancelik/herdr/discussions/1571)

## Summary by CodeRabbit

* **New Features**
  * Added support for resuming sessions with the `herdr:agy` agent.
  * Session resume now recognizes `herdr:agy` session references and automatically forwards the conversation ID when launching the agent.
* **Bug Fixes**
  * Improved agent session snapshot detection to reliably infer and persist the correct `herdr:agy`/`agy` session details when no prior persisted session is available.
* **Tests**
  * Updated automated checks to cover reserved `herdr:agy` handling, expected resume argv, and conversation ID inference/parsing behavior.


ogulcancelik:
Thanks for the contribution. I moved this PR back to draft because the current change only adds the final resume-command mapping. Herdr recognizes and screen-detects AGY today, but it does not have an AGY integration that captures and reports the conversation ID, so the new planner has no production input.

Please implement the AGY integration end to end. AGY supports plugins and command hooks, and hook payloads expose `conversationId`. The Herdr-managed integration should capture that value, report it through `pane.report_agent_session` with the official `herdr:agy` source, and include install/status/uninstall behavior plus the required `docs/next` updates.

Please also validate this against the real AGY CLI rather than only the planner unit test: install the integration, start an AGY conversation, confirm Herdr persists its session reference, restart/restore Herdr, and verify that `agy --conversation <id>` resumes the same conversation. Cover the supported platforms and document any AGY version requirement.

ogulcancelik (changes requested):
please implement the agy integration through its plugin and hook system rather than inferring session ids by scanning `~/.gemini/antigravity-cli/brain`. selecting the newest directory is not robust and can associate multiple panes with the same or wrong conversation.

please use the existing integrations as references. `src/integration/assets/grok/herdr-agent-state.sh` shows a session-only hook reporting an id through `pane.report_agent_session`, while the opencode and kilo integrations show plugin-based integration patterns. the registry, install, status, and uninstall wiring lives under `src/integration/`.

the agy hook should report the exact `conversationId` for its pane using the official `herdr:agy` source. please include install/status/uninstall coverage, `docs/next` updates, and validation against the real agy cli.

Issue #1862, opened 2026-07-25
Headless server CPU remains high with multiple attached clients and working agents


we11adam:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

With three TUI clients attached to the same session and two Codex panes in the `working` state, the Herdr server consistently consumes a noticeable fraction of one CPU core.

Over ten one-second samples, server CPU usage was:
```
average: 20.1%
minimum: 16.2%
maximum: 23.0%
```
No keyboard or mouse input was made during the measurement. Memory usage remained low, but the server CPU usage was sustained.

### Expected behavior

The server should use substantially less CPU during normal multi-client operation and should not sustain around 20% of one core merely while agents are in the working state.

### Reproduction

1. Start the default Herdr server.
2. Open three terminal tabs or windows and run `herdr` in each, attaching them to the same session.
3. Start two Codex panes and wait until `herdr agent list` reports both as `working`.
4. Stop interacting with the clients.
5. On macOS, sample the server process:
```
server_pid=$(pgrep -f '/herdr server')
top -l 11 -s 1 -pid "$server_pid" -stats pid,cpu
```

6. Ignore the first `top` sample and average the remaining ten samples.

The server remains around 16–23% CPU in this setup.

### Impact

The server remains a consistently high-CPU process during an ordinary multi-client workflow, consuming a noticeable fraction of one core for an extended period.

### Environment

- Herdr version: 0.7.5
- Update channel (stable or preview): stable
- Operating system: macOS 27.0 (Build 26A5388g)
- Hardware: M5 Max
- Terminal: Ghostty 1.3.2-main-+e6e26e165 (tip)
- Shell, if relevant: fish@4.8.1
- Relevant config, if any: Default render and animation settings and no render-related overrides.

we11adam:
### Additional technical findings

  I collected an 8-second sample from the Herdr 0.7.5 server while three TUI clients were attached and two Codex panes were in the `working` state.

  Of 1,066 runnable main-thread samples, 1,043 (97.8%) were inside `HeadlessServer::render_and_stream`. The dominant stack was:

```
HeadlessServer::render_and_stream
  -> render_virtual_with_runtime_registry
  -> ratatui::Terminal::try_draw
  -> render_panes
  -> PaneRuntime::render
  -> ghostty_render_state_row_cells_get
```

[...]

PR #1875, opened 2026-07-26
perf: reuse semantic frame buffers for retained updates

we11adam:
## Summary

- reuse the owned semantic frame baseline for retained animation and PTY updates instead of cloning the full `FrameData`
- validate PTY patches before transferring the baseline, then commit the same allocation only after a successful send
- invalidate a transferred baseline on serialization or queue failure so the existing full-render fallback restores a known-good baseline
- add allocation-reuse and failed-send recovery coverage, plus an ignored release-mode benchmark

This is a follow-up optimization to #1868. It does not change visible rendering behavior, the wire format, or `PROTOCOL_VERSION`.

## Context

#1868 avoids pane and Ghostty rendering for animation-only ticks, but semantic retained updates still cloned the complete cached frame before applying a small patch, serializing it, and replacing the old cache. The server already exclusively owns that semantic baseline, so the clone and subsequent drop are unnecessary.

This change transfers the cached frame into the retained update and restores it to the per-client render state after the frame has been sent successfully. Terminal-ANSI clients keep the existing cloned working frame because their baseline is owned by the diff encoder.

## Performance

Run with:

```text
just perf-retained-frame
```

The release-mode benchmark compares the upstream clone strategy with ownership transfer on a 165x43 semantic frame. Each run uses 15 samples per strategy, adapts each sample to at least 75 ms, and alternates strategy order. The table reports the median of three independent run medians:

| Retained update workload | Upstream clone | Owned baseline | Speedup | Hot-path time reduction |
| --- | ---: | ---: | ---: | ---: |
| PTY, one dirty row (129 cells) | 119.443 µs/op | 47.103 µs/op | 2.54x | 60.6% |
| Animation sidebar (36x43 cells) | 133.870 µs/op | 60.519 µs/op | 2.21x | 54.8% |

Both strategies include patch application and protocol framing/serialization. These figures measure the retained-update hot path only; they do not claim an equivalent reduction in total server CPU usage.

## Correctness and fallback behavior

- geometry, runtime availability, hyperlink intersections, and dirty-patch collection are checked before semantic baseline ownership is transferred
- a successful send commits the same frame allocation as the next baseline
- a full render is deferred after queue saturation; because the transferred baseline is not committed, recovery cannot diff against a frame the client never received
- existing full-render fallbacks remain in place for unsafe app state, graphics, overlays, resize, and missing cached frames

## Testing

- `just ci` with Zig 0.15.2
  - 2930 Rust tests passed
  - 17 integration asset tests passed
  - 12 plugin marketplace tests passed
- `just perf-retained-frame` (three independent runs)

Refs #1862

### DIFF
--- a/justfile
+++ b/justfile
@@ -11,6 +11,10 @@ test:
 test-one filter:
     cargo nextest run --locked "{{filter}}" --status-level fail --final-status-level fail --failure-output final --success-output never
 
+# Compare upstream frame cloning with retained semantic baseline ownership transfer
+perf-retained-frame:
+    cargo test --release --locked --bin herdr server::headless::tests::retained_frame_ownership_benchmark -- --ignored --exact --nocapture --test-threads=1
+
 # Run fast local lint checks
 lint:
     cargo fmt --check
--- a/src/server/headless.rs
+++ b/src/server/headless.rs
@@ -3444,7 +3444,7 @@ impl HeadlessServer {
             if !matches!(mode, ClientConnectionMode::App) {
                 continue;
             }
-            let Some(client) = self.clients.get(&client_id) else {
+            let Some(client) = self.clients.get_mut(&client_id) else {
                 return false;
             };
             if client.deferred_render() != DeferredRender::None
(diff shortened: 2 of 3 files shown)


ogulcancelik:
i'm refactoring client-server rendering and especially sidebar atm, but ty!

PR #1877, opened 2026-07-26
fix: preserve focus when closing a non-focused workspace

yianL:
## Context

In herdr `v0.7.5`, closing a workspace through `herdr workspace close <id>` or closing a non-active entry from the sidebar will bring the focus to whichever workspace slid into the closed slot, instead of leaving focus where the user was.

## Changes

- Capture the focused workspace's stable `id` before removal, then use the `id` to restore the `selected`/`active` to its new index.
- New test: `close_non_focused_workspace_keeps_focus` — `[a,b,c]`, focused on `a`, close middle `b` -> focus stays on `a` (previously jumped to `c`)

## Test plan

- `just ci` ✅
- Built the binary locally and manually validated the behavior fix.
  - Create workspace A, B , C.
  - Focus on workspace A.
  - Use the side panel to close workspace C.
  - Assert that we're still focused on workspace A.

refs #1328

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -1630,6 +1630,10 @@ impl AppState {
                 crate::logging::workspace_closed(&workspace_id);
             }
         }
+        let active_workspace_id = self
+            .active
+            .and_then(|idx| self.workspaces.get(idx))
+            .map(|ws| ws.id.clone());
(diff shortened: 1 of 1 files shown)


ogulcancelik:
lgtm ty!

Issue #1789, opened 2026-07-23
Nested Codex process replaces the pane's saved session ID


Pimpmuckl:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

The nested Codex process inherits the pane's Herdr environment and reports its own `SessionStart`.

Herdr then replaces the pane's saved Codex session ID with the nested process's ID. After restarting, Herdr tries to resume that ID instead of the Codex TUI that owns the pane:

```text
ERROR: No saved session found with ID 019f8b52-0b5f-7663-b9cb-945335458338. Run `codex resume` without an ID to choose from existing sessions.
```

I reproduced this in an isolated `herdr-dev` server and traced the original incident to the same trigger: A Codex subagent launched `codex exec --ephemeral`, and the nested session ID replaced the owning TUI session ID.

This is not specific to `--ephemeral`. Any nested Codex process that emits `SessionStart` can overtake the pane's saved session ID. Ephemeral mode just makes the problem obvious because its session cannot be resumed. A persisted nested session could instead make Herdr resume the wrong conversation.

### Expected behavior

Herdr should retain the session ID belonging to the Codex TUI that owns the pane. Codex processes launched underneath that session should not replace it.

### Reproduction

1. Start Codex inside a Herdr pane.
2. Ask that Codex session to run another Codex process:

    `codex exec --ephemeral "Reply only NESTED_OK."`

3. Wait for the nested command to finish.
4. Restart Herdr and reattach the session.

### Impact

Restarting or restoring a Herdr session after such an id takeover can either:

- fail because the saved ID belonged to an ephemeral nested process, or
- resume the wrong Codex conversation if the nested process was persisted.

Note that this doesn't sound like a big deal as manual `codex exec` invocations are rare, but in case of a failed `spawn_agent` tool, GPT agents love pulling that card out of the hat as fallback.

### Environment

- Herdr version: herdr 0.7.5-preview.2026-07-21-0f10e1453a7f
- Update channel (stable or preview): preview
- Operating system: Windows 11
- Terminal: Windows Terminal
- Shell, if relevant: Powershell 7
- Relevant config, if any:

PR #1901, opened 2026-07-26
fix: keep nested codex sessions from replacing pane owner

Pimpmuckl:
## Summary

- keep the pane's owning Codex session when a nested Codex process reports `SessionStart`
- pass the inherited parent thread ID through the Codex integration hook
- clear inherited `CODEX_THREAD_ID` when Herdr launches a new pane shell

## Testing

- `cargo test --locked --bin herdr codex_lifecycle_session_ref_replaces_existing_session_ref`
- `cargo test --locked --bin herdr codex_v2_integration_status_is_outdated`
- Windows PowerShell hook behavior check
- `cargo clippy --bin herdr --locked -- -D warnings`
- `cargo fmt --check`

Native `just ci` currently reaches Unix-only integration targets on Windows; GitHub CI will run the full suite.

Refs #1789

### DIFF
--- a/docs/next/api/herdr-api.schema.json
+++ b/docs/next/api/herdr-api.schema.json
@@ -2833,6 +2833,12 @@
             "pane_id": {
               "type": "string"
             },
+            "parent_agent_session_id": {
+              "type": [
+                "string",
+                "null"
+              ]
+            },
             "seq": {
               "format": "uint64",
               "minimum": 0,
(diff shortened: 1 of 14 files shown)


Pimpmuckl:
Ahh sorry, I typed wrong upstream vs origin on my gh pr open

Sorry for the annoyance!

Issue #1332, opened 2026-07-11
Terminal left in mouse-tracking + kitty-keyboard mode after an abrupt remote-session disconnect


graiz:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

When a remote herdr session (started over ssh) drops abruptly — network loss, laptop sleep, or the ssh transport getting killed — my local Ghostty is left with mouse tracking and the kitty keyboard protocol still on. Moving the mouse spews escape sequences into the shell, keys get misreported. Quitting herdr cleanly restores everything fine; only abrupt drops leave it stuck. This happens most often with a network drop. 

Using claude I made a script to test this and a capture of a dropped session shows these enabled with no matching disable anywhere in the stream: mouse `?1000h ?1002h ?1003h ?1006h ?1015h`, focus `?1004h`, bracketed paste `?2004h`, alt screen `?1049h`, and kitty keyboard `ESC[>7u`.

### Expected behavior

After a remote session ends for any reason — including an abrupt link loss, not just a clean quit — the local terminal should be back to normal: no mouse reporting, normal key handling, and back on the main screen.

### Reproduction

1. From a local Ghostty, `ssh` into a remote host and start herdr (offered/launched remotely).
2. Kill the connection abruptly instead of quitting herdr — e.g. from another local terminal: `pkill -f <remote-host>`, or drop the network / sleep the laptop.
3. Back at the local shell, move the mouse and press a few keys.

### Impact

Every dropped remote session leaves the terminal unusable until I close the tab or manually reset it (printf the disable sequences + `stty sane`). With a flaky link or laptop sleep this happens often.

### Environment

- Herdr version: 0.7.3
- Update channel (stable or preview): stable
- Operating system: macOS 26.0.1 (Apple Silicon / arm64)
- Terminal: Ghostty 1.3.1
- Shell, if relevant: zsh 5.9
- Relevant config, if any: remote session started over ssh; local install at ~/.local/bin/herdr

graiz:
fields were in the post - but I reformatted the comment so it's easier to see. 

Herdr version: 0.7.3
Update channel (stable or preview): stable
Operating system: macOS 26.0.1 (Apple Silicon / arm64)
Terminal: Ghostty 1.3.1
Shell, if relevant: zsh 5.9
Relevant config, if any: remote session started over ssh; local install at ~/.local/bin/herdr

[...]

Issue #939, opened 2026-07-02
SGR mouse sequences leak into pane output after SSH disconnect + reconnect


pWoLiAn:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

After an abrupt SSH disconnect (network drop), moving the mouse on reconnect causes raw SGR mouse escape sequences to appear as literal text inside the focused pane. Example output visible in the
 TUI:

```
 [<5;43;26M[<5;44;26M[<5;45;26M
```

### Expected behavior

Mouse events are consumed by herdr. Nothing leaks into pane output.

### Reproduction

 1. Connect via SSH: `nu.exe → ssh → remote bash → herdr`
 2. Open a TUI inside a herdr pane (e.g. pi)
 3. Drop the network connection (abrupt SSH disconnect — not a clean exit)
 4. Reconnect SSH, reopen herdr, open a pane with any TUI
 5. Move the mouse

### Impact

Every mouse movement produces garbage characters in the active pane. The TUI is unusable until the terminal is manually reset with `printf '\033[?1006l\033[?1003l\033[?1000l'`

### Environment

 - Herdr version: 0.7.1
 - Update channel: stable
 - Operating system: Windows (client), Linux (server). herdr installed on server.
 - Terminal: Windows Terminal
 - Shell: bash (remote)

leandrorazevedo:
Same problem here. It doesn't need a TUI; it happens directly in bash.
You just need to use the mouse scroll wheel.

Connect via SSH: ssh → remote bash → herdr ->  ;18M5;74;18M5;74;18M;18M5;74;18M5;74;18M

takenotice:
I am on v0.7.3 and still get this bug after SSH connection drops. Reproduced on my setup:

- Windows Terminal, WSL2 Debian
- SSH to a Debian VM on LAN, with fast keepalive settings: `ssh -o ServerAliveInterval=5 -o ServerAliveCountMax=2 <host>`
- Run `herdr`
- Physically disconnect the Ethernet cable
Symptoms: mouse movement streams raw SGR sequences into the pane ([<5;43;26M...), and scroll wheel acts like arrow-key history recall instead of scrolling. 

[...]

PR #1902, opened 2026-07-26
fix: clear host mouse reporting after leaving the alternate screen (#1713 / #1332)

### LINKED ISSUE #1713: SGR mouse sequences still leak after SSH disconnect (follow-up to #939)

### Current behavior

After `herdr --remote <target>` exits due to SSH broken pipe (network disconnect), moving the mouse in the terminal produces continuous SGR escape sequence garbage: `[<35;98;31M35;98;30M...`

Button code 35 confirms mode 1003 (any-event tracking) + mode 1006 (SGR encoding) are still active on the host terminal after herdr has exited.

This is a follow-up to #939 — the fix in v0.7.2 added `clear_host_mouse_reporting()` but it runs before `ratatui::restore()` leaves the alternate screen. Some terminals restore saved private mode state (including mouse tracking) on `LeaveAlternateScreen`, re-enabling mouse tracking on the normal screen.

### Expected behavior

Mouse tracking should be fully disabled after herdr exits, regardless of how it exited (clean exit, SSH broken pipe, panic). No SGR mouse sequences should appear in the terminal.

### Reproduction

1. `herdr --remote <target>`
2. Open a TUI with mouse support (e.g. pi) inside a herdr pane
3. Close laptop lid or disconnect network — SSH breaks with "Broken pipe"
4. Re-open terminal — moving the mouse produces SGR garbage characters

### Impact

Terminal becomes unusable after SSH disconnect — every mouse movement produces garbage output. Requires manually running `printf '\e[?1000l\e[?1003l\e[?1006l'` or `reset` to recover.

### Environment

- Herdr version: v0.7.5 (main branch, commit ef85fa0)
- Update channel: preview
- Operating system: macOS (aarch64-apple-darwin)
- Terminal: Ghostty
- Shell, if relevant: zsh
- Relevant config, if any: `herdr --remote` mode with TUI apps using mouse capture in panes

IlyaMichlin:
## Problem

After a remote/SSH disconnect (e.g. `herdr --remote <host>`), the local terminal is left in mouse-reporting mode and emits stray SGR mouse sequences (e.g. `35;22;52M`) on every mouse movement. Recovery requires running `reset`.

## Root cause

`clear_host_mouse_reporting()` ran *before* `ratatui::restore()` left the alternate screen. Terminals like Ghostty track DEC private modes **per screen buffer**, so leaving the alt screen on `LeaveAlternateScreen` restores the primary screen's still-enabled mouse-tracking state — undoing the pre-restore clear, with nothing clearing it afterwards.

This is a follow-up to #939, which added the clear but placed it before the alt screen was left (wrong ordering).

## Fix

Re-clear host mouse reporting immediately *after* each `ratatui::restore()` at all three teardown sites:

- `client::restore_terminal_state` (`src/client/mod.rs`) — the function reached via `TerminalGuard::Drop`
- the panic hook in `src/main.rs`
- the normal teardown in `src/main.rs`

The existing pre-restore clears are kept (the disable sequence is idempotent). The resulting order in `restore_terminal_state` is now: pre-restore clear -> `ratatui::restore()` (leaves alt screen) -> **post-restore clear** -> postlude, so the disable sequence lands on the primary screen after its saved mouse state is restored.

## Testing

- Build passes (`cargo build --bin herdr`).
- Full suite: **2851 pass**. A set of `detect::manifest*` tests (e.g. `detect::manifest::tests::local_override_shadows_cached_remote_manifest`, `detect::manifest_update::tests::process_agent_manifest_commits_newer_manifest_atomically`) is flaky under parallel execution due to shared global cache/config state — the exact failing set is non-deterministic across runs, all pass in isolation / single-threaded, and the **same tests fail identically on the pristine `master` tree**, so they are pre-existing and unrelated to this change.
- Added regression test `terminal_modes::tests::second_clear_after_restore_reemits_full_disable_sequence`, which asserts the pre/post-restore double-clear re-emits the full disable sequence (including SGR mode `1006l`, the source of the stray reports).

## Scope note

This fixes the `herdr --remote` teardown path (the path #1713 reports): the remote thin client sets up its terminal via `setup_terminal` -> `TerminalGuard`, whose `Drop` (and the client panic hook) call `restore_terminal_state()` — the function fixed here.

The `ssh -t <host> 'herdr agent attach ...'` direct-attach-over-plain-SSH path has a **separate root cause**: on abrupt disconnect the remote process receives `SIGHUP` and dies before any `Drop`/panic-hook teardown runs, so no cleanup fires at all. Fixing that requires a signal handler and **will be addressed in a separate follow-up PR**.

Because the `agent attach`/SIGHUP path remains, this does not fully close #1713. **Addresses the `--remote` path of #1713 and #1332; the `agent attach`/SIGHUP path will follow in a separate PR.**

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -595,6 +595,12 @@ fn restore_terminal_state(
     }
 
     ratatui::restore();
+    // Clear host mouse reporting again *after* leaving the alternate screen.
+    // Some terminals (e.g. Ghostty) track DEC private modes per screen buffer
+    // and restore the primary screen's saved mouse-tracking state on
+    // LeaveAlternateScreen, re-enabling it after the pre-restore clear above.
+    // See issue #1713 (follow-up to #939).
+    let _ = crate::terminal_modes::clear_host_mouse_reporting(&mut io::stdout());
     let _ = write_terminal_restore_postlude(&mut io::stdout(), reset_host_color_scheme_reports);
 
     #[cfg(windows)]
--- a/src/main.rs
+++ b/src/main.rs
@@ -774,6 +774,10 @@ fn main() -> io::Result<()> {
         let _ = set_host_color_scheme_reports(false);
         let _ = pop_keyboard_enhancement_flags();
         ratatui::restore();
+        // Re-clear after leaving the alternate screen: some terminals restore
+        // the primary screen's saved mouse-tracking state on
+        // LeaveAlternateScreen, undoing the pre-restore clear. See issue #1713.
+        let _ = crate::terminal_modes::clear_host_mouse_reporting(&mut io::stdout());
         original_hook(info);
     }));
 
@@ -840,6 +844,10 @@ fn main() -> io::Result<()> {
         crate::terminal_modes::clear_host_mouse_reporting(&mut io::stdout())?;
         set_host_color_scheme_reports(false)?;
         ratatui::restore();
+        // Re-clear after leaving the alternate screen: some terminals restore
+        // the primary screen's saved mouse-tracking state on
+        // LeaveAlternateScreen, undoing the pre-restore clear. See issue #1713.
+        crate::terminal_modes::clear_host_mouse_reporting(&mut io::stdout())?;
 
         // Drop app (and all workspaces/panes) before runtime shuts down
         drop(app);
--- a/src/terminal_modes.rs
+++ b/src/terminal_modes.rs
@@ -61,4 +61,41 @@ mod tests {
             );
         }
     }
+
+    // Regression test for issue #1713: after leaving the alternate screen,
+    // teardown clears host mouse reporting a *second* time. Terminals that
+    // track DEC private modes per screen buffer restore the primary screen's
+    // saved mouse-tracking state on LeaveAlternateScreen, so a single
+    // pre-restore clear is not enough. This asserts that calling
+    // `clear_host_mouse_reporting` twice (mirroring the pre-restore and
+    // post-restore calls in the teardown paths) is idempotent and always
+    // re-emits the full disable sequence — including SGR mode 1006, the
+    // source of the stray `35;22;52M`-style reports in the bug report.
+    #[test]
+    fn second_clear_after_restore_reemits_full_disable_sequence() {
+        let mut output = Vec::new();
+
+        // Pre-restore clear (before ratatui::restore leaves the alt screen).
+        clear_host_mouse_reporting(&mut output).unwrap();
+        // Post-restore clear (the fix): must re-emit the disable sequence so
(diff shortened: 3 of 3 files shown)


IlyaMichlin:
Thanks both — good catch, and it's a real limitation. The regression test exercises the `clear_host_mouse_reporting` helper rather than the production restore-ordering, so it wouldn't catch the post-`ratatui::restore()` call being removed or moved back before the alt-screen leave. I called this out in the PR description as well.

The reason it's tested at the helper level rather than the boundary: the three teardown sites (`restore_terminal_state` in `src/client/mod.rs`, plus the panic and graceful paths in `src/main.rs`) all write directly to `io::stdout()`. Asserting the true ordering (clear → leave-alt-screen → clear) would mean refactoring those sites to write through an injectable `impl Write` (and threading a fake through `ratatui::restore()`), which is a non-trivial production change purely for testability — hence I kept the fix minimal.

Happy to do that refactor and add a boundary/integration test if you'd prefer the stronger guarantee — just let me know which direction you'd like. Otherwise I've left the helper-level test in for sequence completeness.

One scope note for reviewers: this fixes the `--remote` teardown path that #1713 reports. The `ssh -t <host> 'herdr agent attach ...'` direct-attach path has a separate root cause (the remote process receives SIGHUP on abrupt disconnect and dies before any Drop/panic-hook teardown runs, so no cleanup fires at all) — that needs a signal handler and will come in a separate follow-up PR.

ogulcancelik:
thanks for investigating this and keeping the patch focused. after tracing the reports, there are two different connection paths being grouped together here.

most reports in #939 and #1332 first ssh into the remote machine and then run herdr. if that ssh transport dies, the remote process can no longer send cleanup sequences to the local terminal. this is a general ssh/terminal limitation that also affects tmux, vim, and similar applications. a remote sighup handler cannot repair the local terminal after the transport is gone. users should prefer `herdr --remote`, mosh, or run `reset` after an abrupt plain-ssh disconnect.

the separate `herdr --remote` claim is locally recoverable in principle because the local herdr client owns the terminal and drops its terminal guard when the ssh bridge closes. we tested that path by killing the exact `remote-client-bridge` ssh process and could not reproduce the leak in ghostty. the proposed explanation also does not match ghostty 1.3.1, where mouse tracking state is terminal-global and leaving the alternate screen does not restore it. the added test only verifies that the reset sequence can be emitted twice; it does not reproduce the reported failure.

i’m closing this pr because it does not address the demonstrated plain-ssh limitation, while the `herdr --remote` change currently lacks a reproducible failure. we can revisit that path if we get a deterministic reproduction showing the local cleanup running but leaving mouse modes enabled.

PR #1925, opened 2026-07-27
fix: make native windows checks reliable

Pimpmuckl:
## Problem

While testing #1909 on native Windows, I ran the repository's local validation commands and encountered an existing Windows-only test flake plus a mismatch between local checks and the Windows CI lane so I had to run the static tests via wslc/docker.

`just lint` and `just check` currently use the Unix recipes on Windows. The resulting `cargo clippy --all-targets` invocation attempts to compile Unix-only integration-test targets and trips unrelated test-only warnings, so the canonical local commands fail before reaching the Windows checks that CI actually supports.

Separately, `windows_shells_round_trip_agent_arguments_through_a_real_command` reads the process-wide `PATH` while integration tests temporarily replace it. Under parallel execution, the shell test can therefore fail to find `powershell.exe` or `cmd.exe`.

## Fix

- Extract the existing inline Windows CI commands into `scripts/windows_check.ps1` with `lint` and `check` modes
- Select native PowerShell recipes for Windows while leaving the Unix recipes unchanged
- Have local Windows checks and Windows CI call the same script
- Allow the generated PowerShell recipes to run under the default Restricted execution policy using a process-scoped bypass
- Serialize the shell round-trip test with the existing integration environment lock

## Scope

On Windows, `just check` intentionally mirrors the currently supported Windows CI lane:

- formatting
- binary-only Windows Clippy
- the `windows_` unit-test filter
- client transport tests
- a Windows build

This does not attempt to make the Unix-only or unfiltered all-target test suites work on Windows. It does not change production or runtime behavior.

## Validation

- `just check` passes on native Windows: Clippy, 123 Windows tests, 21 client transport tests, and build
- The PowerShell recipes pass under Windows PowerShell 5.1 and PowerShell 7
- The recipes launch successfully from a parent process using Restricted execution policy
- The shell round-trip test passed 10 consecutive focused runs
- The shared script preserves the existing Windows CI commands and Zig-cache recovery behavior

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,33 +141,7 @@ jobs:
       - name: Run Windows checks
         if: matrix.kind == 'windows'
         shell: pwsh
-        run: |
-          $ErrorActionPreference = "Stop"
-          function Invoke-Checked {
-            param([string]$Command, [string[]]$Arguments)
-            & $Command @Arguments
-            if ($LASTEXITCODE -ne 0) {
-              throw "command failed with exit code $LASTEXITCODE`: $Command $($Arguments -join ' ')"
-            }
-          }
-          function Invoke-CargoWithZigCacheRecovery {
-            param([string[]]$Arguments)
-            & cargo @Arguments
-            if ($LASTEXITCODE -eq 0) {
-              return
-            }
-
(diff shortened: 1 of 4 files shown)


ogulcancelik:
ty!

PR #1927, opened 2026-07-27
fix: ignore nested codex session reports

Pimpmuckl:
## Problem

Herdr's Codex `SessionStart` hook currently accepts every reported session ID as the pane owner. That is wrong in two Codex cases:

- a nested Codex process inherits the pane's Herdr integration and can replace the parent TUI's saved ID
- `/side` forks a distinct ephemeral Codex thread in-process and also emits `SessionStart`

With GPT 5.6 Sol frequently spawning subagents, this became much easier to encounter in normal use.

After a restart, Herdr may then run `codex resume` with an ephemeral ID that was never persisted, producing `No saved session found`, or with a persisted nested ID and resume the wrong conversation.

Refs #1789.

## Fix

Only record a Codex `SessionStart` when:

- it has a non-empty `transcript_path`, meaning Codex created local persisted state that `codex resume` can use
- an inherited `CODEX_THREAD_ID` is absent or matches the reported ID

Herdr also removes any outer `CODEX_THREAD_ID` when launching a pane. That establishes the pane as a fresh ownership boundary, while Codex processes launched underneath it still inherit the owning thread ID and are rejected when they report a different one.

## Scope

This is intentionally Codex-specific. It does not change shared session storage or other agent adapters.

- normal persisted Codex startup, resume, `/new`, `/fork`, `/clear`, and compaction remain reportable
- `/side` and `--ephemeral` sessions are ignored because they cannot be resumed
- persisted nested Codex processes are ignored even though they have transcript files

The Codex integration version is bumped so existing installs receive the updated hook.

## Validation

- Windows PowerShell and POSIX hook probes report a persisted root session and suppress ephemeral and nested session reports
- a focused pane-launch test verifies that an outer `CODEX_THREAD_ID` is removed
- `just check` passes on Windows
- Local review using GPT 5.6-Sol medium reviewers green


Pimpmuckl:
Fixed in 53046bc0. The hook test now covers the matching inherited `CODEX_THREAD_ID == session_id` allow path and asserts that the session report is preserved.

ogulcancelik:
ty

Issue #1387, opened 2026-07-13
In copy mode arrow keys are not repeated


VincentBaijot:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

In copy mode, holding down an arrow key triggers only a single movement event, forcing the user to press the key repeatedly to move the cursor. By contrast, using the h, j, k, and l keys works as expected, triggering continuous movement while the key is held

### Expected behavior

Arrow keys should behave consistently with alphanumeric movement keys (hjkl). When held down, the cursor should continue to move until the key is released.

### Reproduction

Launch herdr.
Enter Copy Mode (default: Ctrl-b then [).
Attempt to move the cursor by pressing and holding an arrow key.

### Impact

This creates an inconsistent and frustrating user experience. It breaks the expected input behavior for standard navigation keys, significantly slowing down text selection and cursor movement. It is counter-intuitive for users who rely on arrow keys for navigation, as it forces an inefficient, high-effort input pattern compared to the vim-style hjkl navigation.

### Environment

- Herdr version: 0.7.3
- Update channel (stable or preview): stable
- Operating system: Linux (Ubuntu 22.04)
- Terminal: Alacritty 0.16.1
- Shell, if relevant:
- Relevant config, if any:

VincentBaijot:
Ok to mark it as duplicate but @akbash-bot closed both this issue and #1341. Maybe one need to be be open. 

Anyway i will work with one patched version of herdr now because globally its awesome. My patch is (not clean i think) : 

```diff --git a/src/app/mod.rs b/src/app/mod.rs
index 1a0c77a..fa700d0 100644
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1556,6 +1556,8 @@ impl App {
                                 && !self.suppressed_repeat_keys.contains(&key_id)
                             {
                                 self.handle_terminal_key_headless(key);
+                            } else {
+                                self.handle_non_terminal_key_headless(key);
                             }```

Issue #1341, opened 2026-07-12
Copy mode: allow held-key repeat for cursor-movement keys (keep suppressing confirm/exit repeat)


markie1006:
### Summary

In **copy mode**, held cursor-movement keys (arrows, `h/j/k/l`, `w/b/e`, …) do **not** auto-repeat — one keypress moves the cursor exactly one step, even while the key is held down. The same keys auto-repeat correctly in a normal terminal pane. Please allow held-key repeat for copy-mode **navigation** keys, while keeping repeat suppressed for the confirm/exit keys (as today).

### Environment

- herdr **0.7.3** (stable)
- Host: Linux (WSL2)
- Terminal: **Ghostty on macOS**, connected over SSH to the host running herdr
- Ghostty implements the Kitty enhanced keyboard protocol and emits explicit key-repeat events.

### Steps to reproduce

1. From Ghostty (macOS) → SSH → the host, attach to herdr.
2. In a normal pane, hold an arrow key or Backspace → it repeats continuously. ✅ (so repeat events *are* reaching herdr)
3. Enter copy mode (`prefix + [`).
4. Hold an arrow key (or `j`/`k`, `w`, etc.).

### Expected

While the key is held, the copy-mode cursor keeps moving (and, with a selection active, the selection keeps extending) — as in tmux / vim visual mode / iTerm2 copy mode.

### Actual

The cursor moves exactly one cell per physical press. Holding the key does nothing beyond the first step, so navigating more than a few cells requires many discrete presses.

### Analysis

This looks like an intended-but-over-broad safety measure. The 0.3.2 changelog says held-key repeat was restored in terminal panes *"without letting modal close/confirm key repeats leak into the shell."* That guard is important — but the hazard is specific to **action/confirm/exit keys**, not to navigation:

- Repeating a confirm/exit key (`Enter`/`y` = yank & leave, `q`/`Esc` = quit) is dangerous: the first event leaves copy mode and the remaining repeats **leak into the shell** as input. Suppressing repeat for these is correct.
- Repeating a pure **movement** key (arrows, `h/j/k/l`, `w/b/e`, `{`/`}`) only moves the cursor / extends the selection. It never exits the mode and never leaks. Repeat here is harmless and is the expected UX.

So the current behavior appears to suppress repeat for *all* modal keys, which also kills the (safe, desirable) navigation repeat.

### Proposed behavior

[...]

PR #1972, opened 2026-07-28
fix: handle held-key repeat in modal dispatch

mudlej:
## Current behavior

Holding a key in any non-terminal mode acts once. Hold Down in the keybinds list or the navigator and the selection moves a single row. Hold Backspace in either search box and one character is deleted. Copy mode has the same problem (#1341, #1387).

Reproduces in Ghostty and Alacritty on macOS and Linux.

## Cause

Herdr requests `REPORT_EVENT_TYPES`, so kitty-protocol terminals report auto-repeat as `KeyEventKind::Repeat`. But the `Repeat` arm only reaches a pane or the terminal-mode handler. Nothing routes it into the modal dispatch, so every modal drops it.

Introduced by 10b17a6, which correctly stopped repeats leaking into a pane after a held key changed the mode, but scoped the whole repeat path to terminal mode.

## Fix

- `suppressed_repeat_keys: HashSet<_>` 
- becomes 
- `held_key_repeats: HashMap<_, HeldKeyRepeat>` (`Blocked | InMode(Mode)`)

separating the clipboard interception's deliberate block from a press having been handled in a modal. `Repeat` gains a branch that dispatches to the modal handler when the recorded mode still equals the current one.

## Tests

Eleven new tests in `src/app/mod.rs` covering both input paths, the mode guard, each exclusion, and input-source cleanup. All fail before the change and pass after. The existing repeat-suppression and clipboard Ctrl-C tests are unchanged.

Also checked by hand against 0.7.5 does not move on the repeats, this branch does.

## Checks

`cargo fmt --check`, `cargo clippy --all-targets --locked -- -D warnings`, `just integration-assets-test` and `just plugin-marketplace-test` are clean.

`cargo nextest run --locked` is 3093/3097. The four failures are `auto_detect_default_socket_path_from_config_dir`, `auto_detect_writes_client_and_server_logs_to_separate_files`, `no_session_writes_startup_logs_to_monolith_file` and `server_start_restores_legacy_session_through_api_identity`. All four fail identically on an unmodified checkout of the same commit, and none of them send a key event.

Note: Very cool app!

### DIFF
--- a/src/app/input/clipboard.rs
+++ b/src/app/input/clipboard.rs
@@ -1,7 +1,7 @@
 use crossterm::event::{KeyCode, KeyModifiers};
 
 use crate::{
-    app::{App, InputSourceId},
+    app::{App, HeldKeyRepeat, InputSourceId},
     input::TerminalKey,
 };
 
@@ -46,7 +46,8 @@ impl App {
             return false;
         }
 
-        self.suppressed_repeat_keys.insert((source_id, key.code));
+        self.held_key_repeats
+            .insert((source_id, key.code), HeldKeyRepeat::Blocked);
         true
     }
 }
--- a/src/app/input/terminal.rs
+++ b/src/app/input/terminal.rs
@@ -344,8 +344,7 @@ impl App {
             .collect();
         self.pressed_terminal_keys
             .retain(|(id, _), _| *id != source_id);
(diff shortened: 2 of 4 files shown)


ogulcancelik:
superseeded by #2142

if issue still persists please create an issue first.

PR #1973, opened 2026-07-28
fix(input): encode api backtab for enhanced keyboards

DanielOu1208:
## Summary

- normalize API `BackTab` events to `Tab` + Shift before Kitty keyboard encoding
- keep the legacy `CSI Z` encoding unchanged
- cover both canonical API and physical-key representations

## Testing

- `just lint`
- `cargo nextest run --locked input::encode::tests` (52 passed)
- `just ci` reaches an unrelated `events_subscribe_streams_output_and_agent_status_events` timeout that reproduces on untouched `origin/master`


DanielOu1208:
Closing because this change belongs in the client app rather than upstream Herdr.

PR #1975, opened 2026-07-28
fix(skill): let natural-language Herdr mentions trigger the agent skill

0p9b:
## Current behavior

With SKILL.md installed in Claude Code (and other skill-capable agents), `/herdr` works every time, but natural-language requests that explicitly name Herdr — `use herdr to open 3 panes and orchestrate the tasks`, `let herdr manage the panes` — frequently fail: the agent treats "herdr" as plain text, asks what it is, or goes searching for it instead of invoking the skill.

## Expected behavior

An explicit directive mention of Herdr invokes the skill without the slash command, which the current description already intends ("Use only when the user explicitly mentions Herdr").

## Cause

Slash invocation resolves by skill name and bypasses routing. Natural-language invocation routes on the frontmatter `description` alone, and that line works against itself:

- it leads with restrictions ("Use only when…", "Do not use…") and has no affirmative exemplars for a directive mention to match
- it never mentions "herder", which is how users type it and how speech-to-text transcribes it, so most real mentions have no lexical anchor
- "Requires HERDR_ENV=1" reads as a pre-invocation precondition the router can't evaluate, encouraging investigate-first behavior; the actionable check already lives in the skill body

## Change

Rewrites the description only (skill body untouched): affirmative-first, directive exemplars ("use herdr", "open this in herdr", "let herdr manage the panes", …), alias spellings ("herder", "herdr.dev", "the herdr skill"), an invoke-on-mention instruction, and a single suppression clause for mentions that discuss/compare/install/describe Herdr rather than direct it. Adds `tests/skill_frontmatter.rs` pinning that trigger vocabulary so the description can't regress silently.

## Repro

Fresh Claude Code session inside a Herdr pane, skill installed: `use herdr to open 3 panes and orchestrate the tasks`. Before: usually treated as plain text. After: invokes the skill on par with `/herdr`.

## Notes

- Tradeoff: "herder" is an English word, so descriptive mentions can occasionally trigger; the body's HERDR_ENV gate makes a wrong invocation a one-command abort. The suppression clause bounds this.
- No user-facing app behavior changes, so I believe no docs/next update is needed — flagging per CONTRIBUTING in case you want the agent-skill page to mention the trigger vocabulary.
- Environment: Herdr 0.7.5 stable, macOS (arm64), Ghostty, zsh; skill installed for Claude Code, Cursor, Gemini, Grok, codex.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: herdr
-description: "Control Herdr, a terminal multiplexer for coding agents. Use only when the user explicitly mentions Herdr or asks to use Herdr to inspect or control panes, tabs, workspaces, commands, or another agent. Do not use merely because a task could benefit from a background terminal, delegation, or parallel work. Requires HERDR_ENV=1."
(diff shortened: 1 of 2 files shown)


ogulcancelik:
let's discuss this in a issue i didn't have any problems with agents using herdr and adding example cases of semantic match feels overengineering for a skill.

PR #1977, opened 2026-07-28
fix(sidebar): add group_gap to allow gaps between spaces

albe2669:
To improve the layout in the sidebar this PR adds a new config variable called `ui.sidebar.spaces.group_gap`, it controls the gap within a group (a group being 1 space and it's children) while keeping `row_gap` to be between spaces.

Currently, if you set `row_gap=1` you get:
```
<workspace 1>
<row_gap>
∟<worktree 1>
<row_gap>
∟<worktree 2>
<row_gap>
<workspace 1>
```

After this PR you get:
```
<workspace 1>
∟<worktree 1>
∟<worktree 2>
<row_gap>
<workspace 1>
```

If you set `row_gap=1` and `group_gap=1` you get the same as before this PR, aka:
```
<workspace 1>
<group_gap>
∟<worktree 1>
<group_gap>
∟<worktree 2>
<row_gap>
<workspace 1>
```


ogulcancelik:
thanks, the group-aware spacing change makes sense, but i don’t think we should add another setting for spacing inside worktree groups.

please keep worktree parents and their children packed, and apply the existing `row_gap` only between complete groups or unrelated top-level spaces. remove `group_gap` and its config, docs, and tests, while keeping focused coverage for the revised spacing behavior.

also, the current-behavior example in the pr description is inaccurate: today `row_gap` adds a gap between the parent and first child, but consecutive children are already packed.

ogulcancelik:
Fixed directly on master in commit 7db744ab. Closing this PR as superseded.

PR #1983, opened 2026-07-28
fix: resolve new-pane shell via user database with existence checks

bioform:
## What this fixes
  New panes pick the wrong shell when the server has no `$SHELL`. When Herdr's server starts without `$SHELL` in its environment (for example under a service manager or a container init), a new pane opens `/bin/sh` (dash) instead of your login shell.

  ## Why it happens
  `pane_shell_from` reads `$SHELL`. If it is missing, it drops straight to `/bin/sh`. It never asks the OS who you are, and it never checks that the shell it picked exists.

  ## What this changes
  The pane shell is chosen from an ordered list, taking the first one that is a real, runnable program on the host:
  1. the `default_shell` config value (kept as-is — your manual override)
  2. `$SHELL`
  3. your login shell from the OS user database (`getpwuid_r`)
  4. `bash` on `PATH`
  5. `/bin/sh` (last resort)

  Steps 2–4 are existence-checked, so a stale `$SHELL` or passwd entry is skipped instead of breaking the pane. The user-database lookup lives behind `platform::login_shell` and returns `None` on non-Unix.

  ## Tests
  Unit tests cover the order and the existence checks (config wins; `$SHELL` used when installed; skipped when missing; login shell next; then bash; then `/bin/sh`). The lookup is injected, so tests never touch the real environment or filesystem.

  ## Scope
  2 files, ~271 changed lines. No new dependencies. No protocol change.

### DIFF
--- a/src/pane.rs
+++ b/src/pane.rs
@@ -1263,10 +1263,48 @@ fn truncate_handoff_history(history: String, max_bytes: usize) -> String {
 }
 
 fn pane_shell(configured_shell: &str) -> String {
-    pane_shell_from(configured_shell, std::env::var("SHELL").ok())
+    let resolve = |candidate: &str| resolve_shell_for_login_mode(candidate).ok();
+    // `default_shell` is honored verbatim (the always-wins override), but warn
+    // when it does not resolve so a stale config is attributable rather than
+    // surfacing later as an opaque spawn failure in non-login mode.
+    let configured = configured_shell.trim();
+    if !configured.is_empty() && resolve(configured).is_none() {
+        warn!(
+            shell = configured,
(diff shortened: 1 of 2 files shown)


ogulcancelik:
thanks for the contribution. i’m closing this because the current behavior is intentional and documented: new panes use `terminal.default_shell`, then `$SHELL`, then `/bin/sh`. changing that order to consult the user database and prefer `bash` is a feature and policy change, not a bug fix, so it needs discussion and approval first.

the implementation also leaves coderabbit’s major review finding unresolved: the raw `getpwuid_r` behavior belongs in the platform-specific files, not `src/platform/mod.rs`. the later green review skipped the files as similar, so it did not clear that finding.

please open a discussion if you want to propose changing the shell fallback policy.

PR #1984, opened 2026-07-28
fix: count ptmx master fds under devpts layouts

bioform:
## What this fixes
  The test `live_server_holds_one_pty_master_fd_per_pane` fails on WSL2 and many containers.

  ## Why it happens
  The test counts a server's PTY master fds by reading `/proc/<pid>/fd` and keeping only links that resolve to exactly `/dev/ptmx`. On devpts setups such as WSL2, a master fd resolves to `/dev/pts/ptmx` instead, so the test counts zero masters and fails with "had 0 ptmx master fds; expected 1" even though the server holds the fd.

  ## What this changes
  Match both master node paths: `/dev/ptmx` and `/dev/pts/ptmx`. PTY slaves still resolve to `/dev/pts/<N>` and are not counted, so the one-master-per-pane check is unchanged.

  ## Tests
  Test-only change. Verified on a WSL2 host with `cargo nextest --retries 3`: the full suite passes (the previously failing test now passes; one unrelated live test is flaky
  and recovers on retry).

  ## Scope
  1 file, 5 changed lines.

### DIFF
--- a/tests/live_handoff.rs
+++ b/tests/live_handoff.rs
@@ -414,7 +414,8 @@ fn server_ptmx_fd_count(pid: u32) -> usize {
     entries
         .filter_map(Result::ok)
         .filter_map(|entry| fs::read_link(entry.path()).ok())
-        .filter(|target| target == Path::new("/dev/ptmx"))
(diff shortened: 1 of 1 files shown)


ogulcancelik:
ty

Issue #1850, opened 2026-07-25
Navigate mode exits after the first pane-focus movement


Angel-O:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

After entering navigate mode with `prefix+w`, the first `h/j/k/l` pane movement changes the focused pane but also exits navigate mode. Subsequent navigation keys are sent to the newly focused pane.

First raised in #1482 by @akagenorobin. Reopening it here using the standard bug-report format.

### Expected behavior

Navigate mode should remain active after moving between panes. It should exit only when explicitly cancelled, such as with `Esc` or the prefix key.

### Reproduction

1. Create three vertically stacked panes.
2. Focus the bottom pane.
3. Press `prefix+w`.
4. Press `k`; focus moves to the middle pane.
5. Press `k` again.
6. Observe that focus does not move to the top pane and `k` is instead sent to the middle pane's terminal.

### Impact

Navigate mode is effectively limited to one pane movement. Moving through several panes requires pressing `prefix+w` before every movement, which makes persistent navigation unavailable.

### Environment

- Herdr version: 0.7.5
- Update channel (stable or preview): stable
- Operating system: macOS, Darwin 24.5.0 arm64
- Terminal: Warp
- Shell, if relevant: zsh
- Relevant config, if any: no overrides for navigation bindings

we11adam:
@ogulcancelik I reproduced this issue and have a focused one-file fix with a regression test ready. Would you approve a PR from me @we11adam?

PR #1993, opened 2026-07-28
fix: preserve navigate mode after pane focus

we11adam:
## Summary

- preserve Navigate mode after directional pane-focus commands from Navigate keybindings
- keep Prefix-mode pane focus one-shot
- add a regression test through the key-handling path

## Why

The shared pane-focus path settles the app into Terminal mode for one-shot callers. Navigate-mode directional focus reused that path, so the first movement also exited Navigate mode.

The fix restores Navigate mode only when the action originated in Navigate context, leaving the shared API and Prefix behavior unchanged.

## Validation

- `just check`
- manually verified with the patched client and server in an isolated named session

No documentation update is needed because this restores the intended behavior.

Refs #1850

### DIFF
--- a/src/app/input/navigate.rs
+++ b/src/app/input/navigate.rs
@@ -335,11 +335,17 @@ impl App {
                     super::modal::open_rename_pane(&mut self.state, pane_id);
                 }
             }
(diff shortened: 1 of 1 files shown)


ogulcancelik:
ty!

PR #1994, opened 2026-07-28
fix: allow windows-native python command in commit hook

Pimpmuckl:
## Problem

During Windows testing and development, the commit-message hook repeatedly caused hiccups because it invokes `python3`. On Windows, `python3` can resolve to the Microsoft Store alias even when a working Python installation is available through `python`, so commits fail before the validator can run.

## Fix

Use `python` when the hook runs under Git for Windows (`OS=Windows_NT`) and keep `python3` everywhere else.

## Validation

- valid and invalid commit messages exercise the hook correctly in Git Bash on Windows
- the non-Windows branch continues to invoke `python3`
- `just check` passes on Windows

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -4,4 +4,8 @@ set -euo pipefail
 repo_root="$(git rev-parse --show-toplevel)"
 cd "$repo_root"
 
-python3 scripts/conventional_commits.py --message-file "$1"
+if [[ "${OS:-}" == "Windows_NT" ]]; then
+    python scripts/conventional_commits.py --message-file "$1"
+else
+    python3 scripts/conventional_commits.py --message-file "$1"
+fi


ogulcancelik:
@Pimpmuckl i merged this is a mistake but reverted it.  this is fundamentally a dependency problem not os. you need to install python3 :)

Pimpmuckl:
> @Pimpmuckl i merged this is a mistake but reverted it. this is fundamentally a dependency problem not os. you need to install python3 :)

Fair enough, valid point.

Python is installed, but installs natively as `python` on windows on PATH vs `python3` on POSIX hence why I thought this makes sense.

Windows quirk. No clue why.

At the end of the day, it's your call how far you want to support Windows dev.

@ogulcancelik unsure if a reply tags you, doing a tag just in case.

Issue #1982, opened 2026-07-28
Agents in panes are never detected under gVisor (agent list empty, agent_status stays unknown)


smwbev:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

Agents running in panes are never detected. `claude` is alive in the pane and its TUI is visible, but the agent panel stays empty:

```
$ herdr agent list
{"result":{"agents":[],"type":"agent_list"}}

$ herdr agent explain w1:p1
{"error":{"code":"agent_not_found","message":"agent target w1:p1 not found"}}

$ herdr pane process-info --pane w1:p1
{"result":{"process_info":{"pane_id":"w1:p1","shell_pid":86}}}
```

The pane itself knows about the session — `herdr pane get w1:p1` returns `"agent_session":{"agent":"claude","source":"herdr:claude","value":"f42d1223-..."}` (the same UUID as `claude --resume`) with `"agent_status":"unknown"`.

`herdr integration status` reports `claude: current (v7)`, and `HERDR_ENV=1`, `HERDR_PANE_ID`, `HERDR_SOCKET_PATH` are all present in the agent process environment.

`herdr agent start proba --kind claude --pane w1:p6` on a freshly created tab fails with `agent_pane_busy: agent target pane w1:p6 is not an available shell`.

The same happens with Codex CLI in a pane.

### Expected behavior

An agent running in a pane is detected and listed in `herdr agent list` with an idle/working state, and shows up in the agent panel.

### Reproduction

1. Run herdr inside a gVisor sandbox (Bunny Magic Containers).
2. `herdr integration install claude` — status then reports `claude: current (v7)`.
3. Start `claude` in a pane, wait for its TUI.
4. `herdr agent list` → `{"agents":[]}`, `herdr agent explain <pane>` → `agent_not_found`, pane `agent_status` stays `unknown`.

`HERDR_AGENT=claude claude` makes no difference.

### Impact

The agent panel is the main reason we run herdr, and it is empty in this environment: no agent list, no idle/working states, no agent-based navigation. The agents themselves work fine.

### Environment

- herdr version: 0.7.5 (client and server, protocol 17, `compatible: yes`)
- Update channel: stable
- Operating system: Debian 12 (bookworm) x86_64, in a container on Bunny Magic Containers (workloads run under gVisor)
[...]

PR #2000, opened 2026-07-28
fix: detect pane agents when the terminal foreground group is unavailable

smwbev:
Under gVisor (Bunny Magic Containers) `tcgetpgrp` fails with `ENOTTY` and `tpgid` in `/proc/<pid>/stat` stays `0`, so `foreground_process_group_id` returns `None` and `foreground_job` bails out before it collects anything. No pane ever reports a foreground job: `herdr agent list` stays empty, `pane process-info` returns only `shell_pid`, and pane status stays `unknown` even while Claude Code or Codex is running in the pane.

When the foreground group is unavailable, this falls back to the pane shell's own job table. A shell puts each job in its own process group led by the first process of that job, so the shell's direct children carry one group id per job, and the highest id is the most recently started one. Children left in the shell's own group are skipped because they are not a separate job.

On systems that report `tpgid` the fallback is unreachable, so behaviour outside such sandboxes does not change.

Verified on the affected setup (this branch built for linux-x86_64, Debian 12 under gVisor, Claude Code 2.1.220), with no custom hooks or integrations beyond the stock ones:

- before: `agent list` empty, `agent explain` returns `agent_not_found`, `pane process-info` returns only `shell_pid`
- after: the agent is detected, `process-info` reports `foreground_process_group_id` and the `claude` process, and status follows the manifest live — `idle`, then `working` for the whole turn (`rule: osc_title_working`), then back to `idle`

One trade-off worth naming: without `tpgid` a background job such as `sleep &` counts as foreground in these sandboxes, because there is no way to tell it apart from the real foreground job there.

refs #1982

### DIFF
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -96,8 +96,21 @@ pub(crate) fn available_pane_shell(child_pid: u32) -> Option<String> {
 }
 
 pub fn foreground_job(child_pid: u32) -> Option<ForegroundJob> {
-    let tpgid = foreground_process_group_id(child_pid)?;
-    let members = foreground_process_group_members(child_pid, tpgid)?;
+    let group_id = match foreground_process_group_id(child_pid) {
+        Some(tpgid) => tpgid,
+        // Some sandboxed runtimes never report the terminal foreground group:
+        // under gVisor `tcgetpgrp` fails with ENOTTY and `tpgid` in
+        // /proc/<pid>/stat stays 0, so no pane ever reports a foreground job
+        // and agents are never detected. Fall back to the pane shell's own job
+        // table there.
+        None => newest_child_process_group(child_pid)?,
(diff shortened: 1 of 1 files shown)


ogulcancelik:
thanks for investigating this and validating it under gvisor. i'm closing this pr because the current approach is not safe for herdr's core process-detection path.

the proposed implementation adds a process-tree scanning heuristic whenever normal foreground-process detection fails. agent process detection is one of herdr's most performance-sensitive paths, so i don't want a sandbox-specific workaround to introduce additional scanning, ambiguity, or regression risk into the normal linux path.

i'd reconsider a pr if the restricted environment can be identified deterministically and routed through a fully isolated, bounded detection path with no behavior or performance impact on normal systems. it must not activate from a generic process-detection failure.

Issue #1946, opened 2026-07-27
Sidebar divider and section rule are invisible on rose-pine


dineshmistry:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

On `theme = "rose-pine"` the sidebar has no visible edge. The vertical divider between the sidebar and the pane area, and the horizontal rule between the spaces and agents sections, are both missing.

They are actually being drawn. Capturing the client output shows a full-height `│` at the sidebar edge and a `─` rule between the sections, both emitted with foreground `38;2;25;23;36` (`#191724`) — the same value as the theme's own base background, so they paint invisibly.

Overriding `[theme.custom] surface_dim` to any other value brings both lines back, which is how I narrowed it to that token.

I captured the same two lines on the other built-in dark themes and each one draws them in that theme's own base color: catppuccin `#1e1e2e`, tokyo-night `#1a1b26`, dracula `#282a36`, nord `#2e3440`, gruvbox `#282828`, one-dark `#282c34`, kanagawa `#1f1f28`, vesper `#101010`. rose-pine is the case where it disappears completely for me.

### Expected behavior

The sidebar divider and the section rule should be visible against the panel they sit on.

### Reproduction

1. `config.toml`:
   ```toml
   [theme]
   name = "rose-pine"
   ```
2. Launch herdr. Look at the right edge of the sidebar and the gap between the spaces and agents sections — no lines.
3. Add `surface_dim = "#403d52"` under `[theme.custom]` and run `herdr server reload-config`. Both lines appear.

### Impact

Small but constant. On a full-width terminal there is nothing marking where the sidebar ends and the panes begin, and the two sidebar sections run together.

The `surface_dim` override works around it, but that token also fills the tab bar strip and the active space row, so raising it enough to see the lines visibly tints those two areas as well.

### Environment

- Herdr version: 0.7.5
- Update channel (stable or preview): stable
- Operating system: macOS 26.5.2 (arm64)
- Terminal: iTerm2 3.6.11
- Shell, if relevant: zsh
- Relevant config, if any: `[theme] name = "rose-pine"`, no other overrides

PR #2002, opened 2026-07-28
fix: rose pine "surface_dim" theme colour

brabli:
The issue with the previous colour is that it would vanish into the foreground if the user had a corresponding terminal theme set (Ghostty, Kitty, iTerm2).

This PR swaps the colour out for an alternate palette colour that should be visible even with a matching terminal theme.

refs #1946


ogulcancelik:
@brabli thanks, the rose pine change looks right. before merge, could you please visually verify catppuccin with the selected workspace different from the active workspace? the new `surface_dim` is exactly equal to `surface0`, so those row backgrounds may become indistinguishable.

please also update the semantic-frame characterization hashes that are causing the linux and macos checks to fail. if the catppuccin roles do collapse visually, please either choose a distinct `surface_dim` value or keep this pr scoped to rose pine only.

brabli:
@ogulcancelik Fair point re catppuccin colours being the same. After some additional testing I've concluded that the most sensible thing is to just scope this PR to rose pine; I'll create a separate issue/discussion regarding catppuccin.

ogulcancelik:
@brabli review comments point out that colors are still for rose pine too? did you visually verify that the bg colors look different?

brabli:
@ogulcancelik Visually it looks good from my end, the background colours are distinct after the change.

See the comparison with and without the updated colour (note the line on the right is always visible with the change!)

https://github.com/user-attachments/assets/f7675921-490b-477d-bae5-14f5f6dded68

https://github.com/user-attachments/assets/dab35c6e-1e10-4aa4-b250-7275215f04dc

ogulcancelik:
@brabli ty

PR #2013, opened 2026-07-29
fix: harden terminal cleanup and windows checks

onequark10-lgtm:
## What changed

- Restore terminal modes and cursor state exactly once across early failures and normal client shutdown.
- Reject semantic terminal frames whose cell count does not match their declared dimensions.
- Create session and sound temporary files atomically, with private permissions on Unix.
- Make repository maintenance checks deterministic on Windows, including UTF-8 handling and same-drive Zig caches.

## Why

Some error and shutdown paths could leave terminal state unrestored. Predictable temporary paths could also collide with existing files. On Windows, platform-default text decoding and Zig caches on a different drive prevented the documented full check from completing reliably.

## Impact

Normal Herdr behavior is unchanged. Failure handling is safer, malformed frames are rejected, temporary files cannot silently replace existing files, and the Windows repository check completes consistently.

Public documentation does not need an update because these are internal reliability and development-check fixes.

## Validation

- `just check` on Windows
- 124 Windows-focused tests passed
- 21 server transport tests passed
- 93 maintenance tests passed

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 packaging/windows/licenses/* binary
+*.patch text eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+/.zig-cache/
 /vendor/libghostty-vt/zig-pkg/
 __pycache__/
 *.pyc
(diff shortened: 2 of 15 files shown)


onequark10-lgtm:
@coderabbitai review

PR #2031, opened 2026-07-29
fix: list server live-handoff in cli help and completions

gloiiire:
## Problem

`herdr server live-handoff` is dispatched in `run_server_command` (`src/cli/server.rs:10`) and is a real, working command, but it is absent from `server_command()` in `src/cli/spec.rs`.

The spec drives `-h`/`--help` (via `spec::print_requested_help`, called from `src/cli.rs:76` before dispatch) and shell completion generation. The handwritten `print_server_help()` in `src/cli/server.rs:255` *does* list the command — so the two help paths disagree and the discoverable one is the incomplete one:

```
$ herdr server -h
Commands:
  stop                    Stop the running server
  reload-config           Reload config in the running server
  agent-manifests         Show active agent detection manifests
  update-agent-manifests  Fetch and reload agent detection manifests
  reload-agent-manifests  Reload local agent detection manifest overrides

$ herdr server help
herdr server commands:
  ...
  herdr server live-handoff   hand off live panes to a new local server
  ...
```

Same for completions:

```
$ herdr completion zsh | grep -c live-handoff          # 0
$ herdr completion zsh | grep -c reload-agent-manifests # 5
```

This matters because `live-handoff` is the recovery path you reach for manually when `herdr update --handoff` did not run or did not finish — exactly the moment when `--help` is where you look.

## Change

- Declare `live-handoff` in `server_command()` with its three options, matching `parse_live_handoff_params` (`--import-exe`, `--expected-protocol`, `--expected-version`).
- Add a regression test pinning the server subcommand set and the three options, so the spec and the dispatch table cannot drift apart silently again.
- Document the command in the CLI reference (en/ja/zh-cn) and add a `docs/next` changelog entry.

No behavior change to the command itself — it was already reachable, just undocumented.

## Verification

```
$ herdr server -h
Commands:
  stop                    Stop the running server
  live-handoff            Hand off live panes to a new local server
  reload-config           Reload config in the running server
  ...

$ herdr server live-handoff -h
Options:
      --import-exe <PATH>            Server binary to hand the live panes off to
      --expected-protocol <VERSION>  Refuse the handoff unless the new server speaks this protocol
      --expected-version <VERSION>   Refuse the handoff unless the new server reports this version

$ herdr completion zsh  | grep -c live-handoff   # 5
$ herdr completion bash | grep -c live-handoff   # 2
$ herdr completion fish | grep -c live-handoff   # 9
```

`just check` passes on macOS aarch64: `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `just windows-lint`, 2954/2954 nextest, integration-assets and plugin-marketplace tests, and the 93 maintenance script tests.

### DIFF
--- a/docs/next/CHANGELOG.md
+++ b/docs/next/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Relicensed Herdr from AGPL-3.0-or-later to Apache-2.0.
 
 ### Fixed
+- `herdr server --help` and shell completions now list `herdr server live-handoff`, which was dispatched but missing from the CLI spec.
 - Agent prompts now wait briefly after sending text before pressing Enter, preventing prompts from remaining in agent composers without starting a turn. (#1878)
 - Empty clipboard writes from pane applications no longer erase existing clipboard contents or show a copied confirmation. (#1893)
 - Plain mouse movement no longer triggers continuous full renders while preserving Herdr menu hover and pane application mouse tracking. (#1865)
--- a/docs/next/website/src/content/docs/cli-reference.mdx
+++ b/docs/next/website/src/content/docs/cli-reference.mdx
@@ -85,13 +85,14 @@ compinit
 ```bash
 herdr server
 herdr server stop
+herdr server live-handoff [--import-exe <path>] [--expected-protocol <n>] [--expected-version <version>]
 herdr server reload-config
 herdr server agent-manifests [--json]
 herdr server update-agent-manifests [--json]
 herdr server reload-agent-manifests
 ```
 
-`herdr server` runs the headless server explicitly. Use it for supervised or service-style setups. `reload-config` applies reloadable settings without restarting panes. `agent-manifests` shows the active agent detection manifest sources, cached remote versions, and last remote update results. `update-agent-manifests` fetches remote manifest updates immediately, reloads them into the running server, and prints the updated manifest status; pass `--json` for the raw status response. `reload-agent-manifests` reloads agent detection manifests into the running server after local override edits.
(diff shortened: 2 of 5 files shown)


gloiiire:
CI caught that my premise was wrong, and I want to flag it rather than let this sit.

`tests/cli/surface.rs:421` already asserts the opposite of what this PR does:

```rust
assert!(
    !stdout.contains("live-handoff"),
    "internal server handoff command should not be completed: {stdout}"
);
```

…sitting next to `root_help_hides_explicit_client_command`. So `live-handoff` is deliberately classed as internal, not accidentally missing from the spec. The docs agree — every public mention is `herdr update --handoff` or `--handoff` on remote attach, never `herdr server live-handoff` as a command to type. `check (ubuntu-latest)` is red for exactly this reason, and there is no clap middle ground: `.hide(true)` hides from help *and* completions, so the assertion can't be satisfied while listing it in `-h`.

I missed it locally because `tests/cli.rs` is `#![cfg(not(target_os = "macos"))]`, so the whole CLI integration suite is skipped on macOS — 2954 tests locally vs 3097 on ubuntu. `check (macos-latest)` is green here for the same reason. Worth knowing for other macOS contributors: a green `just check` on macOS does not cover `tests/cli/`.

What sent me down this path is a real, smaller inconsistency that this PR does not address: `print_server_help()` (`src/cli/server.rs:259`) *does* advertise the command:

```
$ herdr server -h            # 5 subcommands, no live-handoff
$ herdr server help          # 7 lines, live-handoff listed
```

So the two help paths disagree about whether the command is internal. Three ways I can take it, your call:

1. **Close this.** Intentional is intentional, and I'll drop it.
2. **Invert it** — remove line 259 so both help paths agree the command is internal. Small, matches the existing intent, no design change.
3. **Leave it as-is** if you think the internal designation is worth revisiting now that live-handoff is the manual recovery path when `herdr update --handoff` didn't run or didn't finish. That's a design change though, so it belongs in a Discussion per CONTRIBUTING.md, not in a `fix:` PR — happy to open one instead.

Defaulting to (1) unless you say otherwise. Sorry for the noise.

ogulcancelik:
handoff is experimental still, i don't want to expose it to cli yet.

Issue #2021, opened 2026-07-29
Resume drops the flags an agent pane was launched with


us:
With `[session] resume_agents_on_restore = true`, a Claude Code pane is restored as `claude --resume=<id>` and the flags it was originally launched with are dropped.

I launch agents as `claude --dangerously-skip-permissions`. That flag is launch-only, there is no config-side workaround: `permissions.defaultMode = "bypassPermissions"` in `~/.claude/settings.json` is rejected on startup.

```
Cannot set permission mode to bypassPermissions because the session was not launched with --dangerously-skip-permissions
```

So after every restart the restored panes start asking for permission again. Same for other agents whose autonomy level is a launch flag only (codex `--full-auto` etc).

Two ways this could work, either is fine:

- reuse the argv the pane was originally started with when resuming
- a config escape hatch:

```toml
[session.resume_args]
claude = ["--dangerously-skip-permissions"]
```

Current workaround is a zsh function shadowing `claude` that re-injects the flag when `--resume` is present, which only works because the resume command goes through the shell.

herdr 0.7.5, macOS 26.5.2 arm64, claude code 2.1.220

Issue #965, opened 2026-07-03
Session restore relaunches agents without their original flags, environment, and pane cwd


kunchenguid:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

After `herdr server stop` and relaunching the same named session, the layout restores great - workspaces, tabs, panes, ids, labels all intact, and each pane's agent relaunches with its session id. But the relaunch drops everything else about how the agent was originally started:

- command-line flags are gone: panes I launched as `claude --dangerously-skip-permissions` came back as plain resumed `claude`, silently downgraded to per-command approval prompts (the missing "bypass permissions on" footer line is the tell)
- environment is gone: `CLAUDE_CODE_ENABLE_PROMPT_SUGGESTION=false` set at launch was not present after restore
- the pane restarts at the TAB's creation cwd, not the directory the pane was actually working in

I hit this across multiple server restarts in one afternoon running a fleet of agent panes. Every pane needed manual repair: interrupt, exit, relaunch with the right cwd + flags + `--resume <id>`.

### Expected behavior

A restored pane relaunches its agent with the same argv, environment, and per-pane working directory it had before the stop. The session-id resume part already works - it's the launch context around it that gets lost.

### Reproduction

1. `herdr`, create a tab in directory A
2. in the pane: `cd /some/other/dir-B`, then `FOO=bar claude --dangerously-skip-permissions`
3. `herdr server stop`, relaunch `herdr`, let the session restore
4. the pane is back and claude resumes with context, but: no `--dangerously-skip-permissions` (default permission mode in the footer), `FOO` unset, cwd is A instead of dir-B

### Impact

Unattended agent fleets silently lose autonomy on every server restart - each agent freezes at its first tool-approval prompt until a human fixes the pane by hand. Bit me twice in one afternoon before I worked out the repair routine.

### Environment

- Herdr version: 0.7.1
- Update channel (stable or preview): stable
- Operating system: macOS 26.5.1 (Apple Silicon, M4 Pro Mac mini)
- Terminal: WezTerm
- Shell, if relevant: zsh
- Relevant config, if any: config.toml managed via nix/home-manager; `update_check = false`, `manifest_check = false`

ogulcancelik:
thanks for the report. i’m going to close this as a feature request rather than a bug.

[...]

PR #2036, opened 2026-07-29
fix: preserve agent launch arguments across restore

IsaiaScope:
## Current behavior

With `resume_agents_on_restore`, a restored pane relaunches its agent as `claude --resume <id>`, dropping the flags the pane was originally launched with. An agent started as `claude --dangerously-skip-permissions` comes back asking for permission on every tool call.

There is no config-side workaround for this particular flag: setting `permissions.defaultMode = "bypassPermissions"` in `~/.claude/settings.json` is rejected at startup with `Cannot set permission mode to bypassPermissions because session was not launched with --dangerously-skip-permissions`. The same applies to other agents whose autonomy level is launch-flag-only, e.g. `codex --full-auto`.

## Expected behavior

A restored pane relaunches its agent with the same arguments it was started with, alongside the existing session-id resume.

## Reproduction

1. `herdr agent start demo --kind claude --pane <id> -- --dangerously-skip-permissions`
2. `herdr server stop`, then relaunch herdr and let the session restore
3. inspect the pane's foreground process: `claude --resume <id>` — the flag is gone

## What changed

The arguments passed after `--` at agent start are recorded on the pane, persisted in the session snapshot next to the existing agent-session identity, and spliced back in at position 1 when the resume command is built.

- `agent_resume::splice_launch_args` inserts the saved args directly after the executable, before the resume flags. Position 1 is the only placement correct for both argv shapes in `plan()` — flag-form (`claude --resume <id>`) and subcommand-form (`codex resume <id>`).
- `PaneAgentSessionSnapshot.launch_args` is `#[serde(default, skip_serializing_if = "Vec::is_empty")]`, so older snapshots load unchanged and panes without extra args serialize exactly as before.
- Args are re-applied to the restored pane so they survive repeated restarts, not just the first one.
- `clear_agent_name` clears them, so a different agent later started in the same pane does not inherit them.

This is separate from the existing `launch_argv`, which describes a pane spawned running a command instead of a shell. A pane is one or the other, and agent panes are shell-backed.

## Scope

The change is agent-agnostic: it applies to all 15 agents that `agent_resume::plan()` supports, since nothing in the splice or the persistence is agent-specific. The remaining 6 agent kinds have no resume plan at all, so there is nothing to preserve for them.

## Verification

- new unit test covers claude and codex, i.e. both argv shapes, plus the empty-args case
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` clean
- `persist` 76/76, `agent_resume` 24/24
- manually verified end-to-end on macOS with claude in an isolated named session: flag present in the live pane process after two consecutive restarts, present in the snapshot after each, and a snapshot with `launch_args` removed still restores and resumes without it

Live end-to-end testing was done with claude only; the other agents are covered by the shared code path rather than by observation.

## Docs

User-facing behavior changes, so a public-doc note may be wanted. No doc files touched here.

Related: #965 (closed, covered pane cwd and environment) and #2021.

## Summary by CodeRabbit

* **New Features**
  * Preserved agent launch arguments when starting, saving, and restoring terminal sessions.
  * Restored agents now receive their original launch arguments before resume options.
  * Empty launch arguments are handled without changing existing behavior.

* **Bug Fixes**
  * Improved session restoration so agent startup configuration remains consistent across saved and resumed sessions.


ogulcancelik:
sorry but this needs proper design first. please don't open prs for unaproved features.

IsaiaScope:
No worries! Thank you very much for your software! Keep going! It was just a way to be helpful

PR #2041, opened 2026-07-29
fix(client): restore terminal on SIGHUP and SIGTERM

MattJColes:
## What

The client only enabled `ctrlc` without its `termination` feature, so it caught SIGINT alone. On a direct SIGHUP or SIGTERM the process died without unwinding, so neither `TerminalGuard::Drop` nor the panic hook ran and `clear_host_mouse_reporting()` never fired. Mouse tracking stayed on and leaked SGR reports as garbage on every pointer move.

This is a direct-signal path, e.g. a terminal emulator SIGHUPing its foreground child on window close. It is **not** the `herdr --remote` ssh-death path. Under `--remote` the local `herdr client` is a subprocess talking to the bridge socket; when ssh dies the client sees EOF on the socket, the loop returns an error, and `drop(terminal_guard)` restores the terminal. No signal is delivered, so that path was already clean.

## Fix

Enable `ctrlc`'s `termination` feature so the existing handler also catches SIGTERM and SIGHUP. It sets `should_quit`, the loop exits, and `TerminalGuard::Drop` restores the terminal the same way it does for Ctrl+C. Handler install failure is logged with `warn!` rather than swallowed. No behaviour change on SIGINT.

## Tests

`tests/client_mode.rs` drives a client under a pty and checks for the mouse teardown bytes emitted after the trigger:

| Test | master | this PR |
|---|---|---|
| `client_restores_terminal_on_server_eof` (kill server → EOF) | pass | pass |
| `client_restores_terminal_on_sighup` (SIGHUP to client) | fail (zero teardown bytes) | pass |

The EOF test passing on master confirms the `--remote`/ssh path already restores. The SIGHUP test is the regression this PR fixes.

## Note

The #1713 SGR leak looks like an ordering issue rather than signals: `restore_terminal_state` clears the mouse modes before `ratatui::restore()` leaves the alt screen, and terminals like Ghostty restore saved private-mode state on `LeaveAlternateScreen`. Not addressed here.

refs #1713


MattJColes:
Good catch, fixed in 66772d1a. I log the failure with `warn!` rather than returning it. `ctrlc::set_handler` fails mainly on double-registration (one handler per process), and aborting client startup for that would be worse than the degraded state, `TerminalGuard::Drop` and the panic hook still cover normal exit and panics. The warning makes the degraded case visible instead of silent.

ogulcancelik:
thanks for the pr. i think this addresses a different path: when ssh dies, the local thin client gets eof through the bridge socket and exits normally; it does not receive sigterm or sighup. your reproduction signals the client directly, so it does not reproduce an `--remote` ssh failure.

i also couldn’t reproduce an sgr leak by killing the ssh bridge. do you have a reproduction where the bridge dies but the thin client fails to restore the terminal?

MattJColes:
Hey, i think we were just talking past each other, youre right. went and wrote two integration tests in `tests/client_mode.rs` that drive a client under a pty and check for the mouse teardown bytes, and they match what you said.

the ssh-death path never hits a signal. `run_remote` spawns a local `herdr client` subprocess talking to the bridge socket, so when ssh dies the client just gets eof on the socket, the loop bails with an error and `drop(terminal_guard)` does the restore. the test that kills the server to force that eof passes on master untouched, so no leak there and my repro in the pr doesnt reproduce the `--remote` case.

where it does apply is a direct SIGHUP/SIGTERM to the client, like when a terminal emulator sighups its foreground child as the window closes. that test fails on master with zero teardown bytes and passes once `termination` is on. so it still fixes something, just not the path #1713 is describing.

one other thing while i was in there. i think the #1713 leak might be ordering, not signals. in `restore_terminal_state` we clear the mouse modes before `ratatui::restore()` leaves the alt screen, and ghostty puts back saved private mode state on `LeaveAlternateScreen`, so mouse tracking comes back on the normal screen.

MattJColes:
done in be1d2cea - shared setup pulled into `assert_client_restores_terminal`, tests just pass their trigger closure now

ogulcancelik:
ty!

PR #2042, opened 2026-07-29
fix: detect pane agents in sandboxes without a foreground process group

smwbev:
Follow-up to #2000, which was closed because the fallback activated on a generic foreground-detection failure and added scanning to the normal Linux path. This reworks it along the lines you asked for there.

The environment is now identified deterministically instead of being inferred from a failure. gVisor names itself in its kernel log banner (`Starting gVisor...`), read once per process through `syslog(2)` and cached in a `OnceLock`. An unreadable or restricted kernel log counts as "not that environment", so the branch stays off by default. This mirrors the existing `running_inside_wsl` marker check in the same file.

The normal path is untouched. When `foreground_process_group_id` returns a tpgid, `foreground_job` behaves exactly as before and never reaches the sandbox branch; a missing tpgid on its own no longer enables anything. On a normal system the added cost is one cached bool read in the case where detection already returned nothing.

The sandbox branch is bounded: it reads the pane shell's children list and inspects at most `SANDBOX_CHILDREN_SCAN_LIMIT` (64) entries, with no process-tree walk. Among the shell's direct children it takes the newest job group and skips children still in the shell's own group, comparing against the shell's real pgrp rather than its pid.

Verified on Bunny Magic Containers (gVisor), Debian 12, with this branch built for linux-x86_64 and no custom hooks or integrations beyond the stock ones:

- before: `agent list` empty, `agent explain` returns `agent_not_found`, `pane process-info` returns only `shell_pid`
- after: agents are detected, `process-info` reports `foreground_process_group_id` and the agent process, and status is applied and kept up to date from the manifest — observed `idle`, `working` and `done` with the state sequence advancing as the pane changed

Tests cover the kernel-log marker parsing, newest-job selection, skipping the shell group when the shell is not its own group leader, and that the scan stops at the limit.

Trade-off, unchanged from the previous discussion: inside such a sandbox a background job counts as foreground, because the information needed to tell them apart does not exist there. Outside the sandbox nothing changes.

refs #1982

### DIFF
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -4,6 +4,7 @@ use std::{
     os::fd::RawFd,
     path::PathBuf,
     process::{Command, Stdio},
+    sync::OnceLock,
 };
 
 use super::{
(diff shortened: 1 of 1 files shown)


smwbev:
Good catch on the buffer size — `SYSLOG_ACTION_READ_ALL` returns the tail of the log, so a 1 KiB read could have cut off the boot banner. Raised the probe to 8 KiB and documented why that is enough.

Measured inside the affected sandbox:

```
SYSLOG_ACTION_SIZE_BUFFER -> 131072      (capacity)
SYSLOG_ACTION_READ_ALL    -> 650 bytes   (the whole gVisor log, 14 lines)
read again 2s later       -> 650 bytes   (no growth)
```

The gVisor log is a fixed banner set rather than a growing kernel log, so it fits well inside the probe, and hosts with a real (larger) kernel log never carry the marker to begin with.

ogulcancelik:
thanks for iterating on this. i realized the kernel-log probe still is not a robust contract, so the safer design is a general explicit opt-in: `HERDR_PROCESS_DETECTION=child-groups`. native detection stays the default, and the alternative path only runs when the server starts with that value.

i implemented that design, including the polling fix, in #2052. i am closing this pr as superseded. thank you for the reproduction, gvisor validation, and initial bounded fallback.

PR #2052, opened 2026-07-30
fix: add opt-in child-group process detection

ogulcancelik:
Restricted Linux runtimes can omit terminal foreground process groups, leaving Herdr unable to identify pane agents. This adds an explicit `HERDR_PROCESS_DETECTION=child-groups` server mode that infers from the pane shell's bounded direct-child scan only when native detection is unavailable.

Native detection remains the default and returns before the opt-in path. Inferred groups are kept out of native foreground-group change tracking, preventing the fallback from triggering process probes every detector tick. Returning the shell group when no separate job exists also restores idle-shell and agent-exit behavior.

Includes next-release docs and tests for mode parsing, group selection, shell fallback, scan limits, and probe scheduling.

Supersedes #2042. Thanks @smwbev for the reproduction, gVisor validation, and initial bounded fallback.

Validation: `just check`

refs #1982

### DIFF
--- a/docs/next/CHANGELOG.md
+++ b/docs/next/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 - Vibe and other Kitty-keyboard pane applications now receive shifted letters and punctuation when they request associated text. (#2020)
+- Linux runtimes without terminal foreground process groups can opt into child-group agent detection with `HERDR_PROCESS_DETECTION=child-groups`. (#1982)
 - Installing the Herdr agent skill with the `skills` CLI no longer copies the entire repository. (#2022)
 - Agent prompts now wait briefly after sending text before pressing Enter, preventing prompts from remaining in agent composers without starting a turn. (#1878)
 - Empty clipboard writes from pane applications no longer erase existing clipboard contents or show a copied confirmation. (#1893)
--- a/docs/next/website/src/content/docs/agents.mdx
+++ b/docs/next/website/src/content/docs/agents.mdx
@@ -51,6 +51,8 @@ Claude Code, Codex, GitHub Copilot CLI, Droid, Qoder CLI, Cursor Agent CLI, and 
 
 On Linux and macOS, a host-visible wrapper can hide the real agent process from Herdr. Set `HERDR_AGENT=<agent>` on the wrapper command to tell Herdr which existing agent screen manifest to use. For example, run `HERDR_AGENT=claude fence -- claude` on Linux or `HERDR_AGENT=claude nono run --profile claude-code -- claude` on macOS. The hint is scoped to that foreground process; setting it only inside a VM or container is not visible to Herdr, and you should avoid exporting it globally unless every inherited foreground process should be treated as that agent.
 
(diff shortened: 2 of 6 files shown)


ogulcancelik:
fixed the lifecycle-authority case in 1d9251dd. when the kernel exposes no foreground group, lifecycle authority no longer suppresses all future process probes; the existing five-second identified-agent interval remains active, while inferred groups still stay out of per-tick change tracking. added a regression test and reran `just check`.

ogulcancelik:
@smwbev could you test this branch on the affected Bunny/gVisor setup before we merge? Start the Herdr server with `HERDR_PROCESS_DETECTION=child-groups` and verify Claude/Codex discovery, idle → working → idle/done transitions, exit back to the shell, and `herdr agent start` from an idle pane. Please also note whether idle CPU changes with several panes. The mode is server-wide, but native detection still wins per pane when available.

smwbev:
Tested `issue/1982-process-detection-mode` (1d9251dd) on the affected setup: Bunny Magic Containers (gVisor), Debian 12 x86_64, built for linux-x86_64 from this branch, Claude Code 2.1.220, Codex CLI 0.145.0. Server started from the pane shell environment with `HERDR_PROCESS_DETECTION=child-groups`, confirmed in `/proc/<server pid>/environ`. No custom hooks or integrations beyond the stock ones.

**Discovery** — both agents are found, which was the original failure:

```
agent list -> [('w5:p1','claude','idle'), ('w4:p8','codex','idle')]
pane process-info --pane w5:p1 -> foreground_process_group_id + the claude process
```

**Transitions** — `idle -> working -> done` observed for both, polling every 2s:

```
codex    idle -> t+2s working -> t+4s done
claude   idle -> t+2s working -> t+4s done
```

**Exit back to the shell** — after quitting Codex the pane cleanly stops being an agent pane:

```
agent get w4:p8 -> agent_not_found
agent list      -> pane gone from the list
pane get w4:p8  -> agent: none, agent_status: unknown
```

**`herdr agent start` from an idle pane** — works now (it failed with `agent_pane_busy` before this branch):

```
agent start proba --kind claude --pane w4:p8
-> {"agent":"claude","agent_status":"idle","interactive_ready":true,"name":"proba",...}
```

**Idle CPU with several panes** — no measurable difference between modes. Fresh session, 4 panes with plain shells and no agents, server CPU delta over 60s:

| mode | panes | agents | CPU over 60s |
|---|---|---|---|
| `child-groups` | 4 | 0 | 1s (1.7% of one core) |
| `native` | 4 | 0 | 1s (1.7% of one core) |

With agents actually detected it rises to 3–4s per 60s (5–6.7%), but that is the state detector doing its normal work per agent, not the child-group scan: the same panes with the agents killed drop straight back to 1s. I did not see per-tick probing.

One note in case it matters for future measurements here: under gVisor `utime`/`stime` in `/proc/<pid>/stat` stay `0` (a 3s busy loop does not move them), so CPU has to be read via `ps -o cputime` instead.

Everything on your list behaves correctly from my side.

ogulcancelik:
ty @smwbev, i think this is the best we can do atm. merged!

Issue #1362, opened 2026-07-12
OpenCode child sessions remain idle in split panes


jackjinke:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

When an OpenCode child session runs in its own Herdr split pane (e.g. multiplexer support from oh-my-opencode-slim), that pane remains idle while the child session is actively working. The parent OpenCode pane can also remain idle while work is running only in a child session. After detaching and reattaching a child session, its pane can continue showing idle during new work.

<img width="1215" height="698" alt="Image" src="https://github.com/user-attachments/assets/b4faef7d-47bf-420b-8c65-5e248b37b6d6" />

### Expected behavior

An attached child session's pane should show working or blocked according to that session's current state. The parent OpenCode pane should show working or blocked while any child session has that state. I have edited the OpenCode plugin locally so that it shows agent status like this:

<img width="286" height="122" alt="Image" src="https://github.com/user-attachments/assets/03b6c0d8-4775-487e-b9d2-5eec5594446a" />

### Reproduction

1. Install the OpenCode integration with `herdr integration install opencode`.
2. Start OpenCode in a Herdr pane.
3. Start a child session using OpenCode's task tool and attach it in a Herdr split pane, for example: `opencode attach http://127.0.0.1:4096/ --session ses_example --dir /path/to/project`.
4. Replace the URL, session ID, and directory with the values for the running OpenCode server and child session, then run work in the child session.
5. Observe that the child pane and parent pane show idle while the child is working.
6. Detach the child session, reattach it in a split pane, and run more work.
7. Observe that the reattached child pane remains idle.

### Impact

The workspace status does not show which OpenCode sessions are active, so I cannot reliably monitor parent and child agents at a glance.

### Environment

- Herdr version: 0.7.3
- Update channel (stable or preview): stable
- Operating system: Linux
- Terminal: Ghostty (Mac)
- Shell, if relevant: zsh
- Relevant config, if any: bundled OpenCode integration version 8

xor-xe:
I hit this too and have a plugin-only fix working locally (child-session state aggregation and routing to attached panes) with tests. Can I get approval to open a pr?

tomjamescn:
[...]

PR #2062, opened 2026-07-30
fix: track opencode child session state

huilang021x:
## Summary

- aggregate child session state without letting a child idle event override an active root session
- treat `opencode attach --session` panes as owners of the attached child session
- ignore stale or foreign session lifecycle state and clear nested child state when its parent is deleted

## Testing

- `bun test src/integration/assets/herdr-agent-state.test.ts`
- `bun test src/integration/assets/opencode/herdr-agent-state.test.ts`

Refs #1362

### DIFF
--- a/src/integration/assets/opencode/herdr-agent-state.js
+++ b/src/integration/assets/opencode/herdr-agent-state.js
@@ -16,6 +16,7 @@ let reportedRootSessionID;
 const sessionParents = new Map();
 const childStates = new Map();
 let ownedRootSessionID;
+let awaitingRootCreation = false;
 let rootState = "idle";
 const CHILD_EVENT_STATES = new Map([
   ["permission.asked", "blocked"],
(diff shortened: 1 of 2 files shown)


huilang021x:
@coderabbitai review

huilang021x:
@coderabbitai review

ogulcancelik:
Thanks for working on this. The attached/resumed-session binding used by this approach cannot currently be implemented reliably in the Herdr plugin. OpenCode does not emit a startup event containing the resumed session ID, and the plugin sees the Bun worker argv rather than the user-supplied `attach --session` arguments. As a result, `attachedSessionIDFromArgv()` cannot observe those arguments in production even though the mocked tests pass. This needs upstream lifecycle support, tracked in https://github.com/anomalyco/opencode/issues/39711, so I’m closing this PR. The parent/child aggregation behavior may still be Herdr-side work, but it should be revisited separately once there is a reliable ownership signal.

Issue #2082, opened 2026-07-30
Pi pane scroll position is lost on terminal resize


drgnchan:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

When a Pi pane has scrollback and its viewport is scrolled up, changing the terminal width moves the viewport back to the live bottom. The transcript is still present after Pi redraws, but the user's scroll position is lost. This reproduces with mobile portrait/landscape changes and repeated width changes.

### Expected behavior

A Pi pane that is scrolled up should preserve its scroll offset across terminal resizes. Repeatedly switching between two widths without scrolling should not move the viewport.

### Reproduction

1. Start Herdr and run Pi 0.83 in a pane.
2. Create or load a Pi session long enough to have pane scrollback.
3. Scroll the pane up with the mouse wheel or PageUp.
4. Change the terminal width, for example by rotating a mobile terminal between portrait and landscape.
5. Observe that the pane viewport moves to the live bottom instead of remaining scrolled up.

### Impact

Resizing or rotating the terminal loses the reader's place in long Pi transcripts, making mobile review unreliable.

### Environment

- Herdr version: 0.7.5
- Update channel (stable or preview): stable
- Operating system: Ubuntu 24.04.3 LTS
- Terminal: iTerm2-compatible terminal over SSH (`LC_TERMINAL=iTerm2`, `TERM=xterm-256color`)
- Shell, if relevant: Bash
- Relevant config, if any: none

ogulcancelik:
this is upstream pi behavior rather than a herdr resize regression. on width changes, pi performs a destructive full redraw that clears scrollback and rewrites the transcript, which invalidates the viewport position herdr preserved during resize.

please see earendil-works/pi#6702 and earendil-works/pi#7194. pi’s new experimental `--alt` renderer addresses this and will be available in the next release.

PR #2086, opened 2026-07-30
fix: preserve Pi scroll position across resizes

drgnchan:
## Summary

- preserve a scrolled Pi viewport while Pi rebuilds its primary-screen transcript after a resize
- allow Pi's scrollback clear to replace the old transcript instead of accumulating duplicate redraws
- restore the saved offset only after synchronized output completes, including redraws split across PTY reads and overlapping resizes
- keep the existing Droid-specific scrollback-clear filtering behavior scoped to Droid

## Testing

- `just ci`
- live Pi 0.83 reproduction in an isolated Herdr session: five 35/53-column round trips kept `offset_from_bottom` at `232`, with stable scrollback size and visible-content hash for each width

Refs #2082

No public documentation update is needed; this restores the expected resize behavior.

### DIFF
--- a/src/pane/osc.rs
+++ b/src/pane/osc.rs
@@ -691,19 +691,8 @@ pub(super) fn current_transient_default_color_owner(shell_pid: u32) -> Option<u3
     (!foreground_job_is_shell(&job, shell_pid)).then_some(job.process_group_id)
 }
 
-fn foreground_job_uses_droid_scrollback_compat(job: &crate::platform::ForegroundJob) -> bool {
-    job.processes.iter().any(|process| {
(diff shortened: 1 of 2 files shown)


ogulcancelik:
please see my comment on your issue

https://github.com/herdrdev/herdr/issues/2082#issuecomment-5132843392

i don't want to bring flaky detection if upstream pi gonan fix it.

drgnchan:
Understood, thanks for the context. I’ll keep this as a local compatibility patch for now and revisit it after Pi releases the experimental alternate-screen renderer.

PR #2092, opened 2026-07-30
fix(omp): support Windows session paths

art-wiedzmin:
## Summary

- accept POSIX and Windows absolute OMP session paths
- retain the session-id fallback for relative or missing paths
- bump the managed OMP integration from v7 to v8

On Windows, `SessionManager.getSessionFile()` returns paths such as `C:\Users\User\.omp\agent\sessions\session.jsonl`. The current `file.startsWith("/")` check rejects them, so Herdr receives `kind: id` instead of `kind: path`.

## Verification

- `bun test src/integration/assets/herdr-agent-state.test.ts --test-name-pattern "OMP accepts POSIX and Windows session paths"` — 1 passed
- `cargo fmt --check` — passed
- `git diff --check` — passed
- live Windows smoke: a newly started Herdr-managed OMP session reported `agent_session.kind = path` with a `C:\...jsonl` value

### DIFF
--- a/src/integration/assets/herdr-agent-state.test.ts
+++ b/src/integration/assets/herdr-agent-state.test.ts
@@ -227,6 +227,17 @@ for (const integration of integrations) {
   });
 }
 
+test("OMP accepts POSIX and Windows session paths", async () => {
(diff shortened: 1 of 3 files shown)


ogulcancelik:
ty!

PR #2108, opened 2026-07-31
docs: add windows beta install via mise

iki:


iki:
@ogulcancelik maybe docs could be allowed in contributors PR's too, WDYT?

Issue #2135, opened 2026-07-31
Nested-launch guard rejects herdr inside a plain tmux session that is not a herdr pane


WSHAPER:
### Is this a reproducible bug?

- [X] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [X] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

Running `herdr` inside a plain tmux session that is not a herdr-managed pane is rejected as a nested launch:

```
$ herdr
error: nested herdr is disabled by default.
see configuration if you want to enable it.
"recursion is a pathway to many abilities some consider to be... unnatural."
```

### Expected behavior

`herdr` should start normally in a tmux session that is not a herdr-managed pane.

### Reproduction

1. From inside a herdr-managed pane, start a detached tmux server:
   ```
   tmux new-session -d -s s1
   ```
2. From anywhere, create a second, unrelated session on that same server:
   ```
   tmux new-session -d -s s2
   ```
3. In a pane of the new session, run `herdr`. It prints the "nested herdr is disabled" error above.

Observed in the unrelated session's pane:

```
$ printf 'HERDR_ENV=[%s] HERDR_PANE_ID=[%s]\n' "$HERDR_ENV" "$HERDR_PANE_ID"
HERDR_ENV=[1] HERDR_PANE_ID=[w7:p1]
```

### Impact

Once any tmux server has been started from a herdr pane, `herdr` cannot be launched in any session on that server unless `allow_nested` is enabled. The only workaround is killing the whole tmux server and restarting it from outside herdr.

### Environment

- Herdr version: 0.7.5
- Update channel (stable or preview): stable
- Operating system: Ubuntu 26.04 LTS (Linux 7.0.0-28-generic x86_64)
- Terminal: tmux 3.6
- Shell, if relevant: bash
- Relevant config, if any: none (default config, no `allow_nested` setting)

ogulcancelik:
thanks for the report. herdr is a terminal multiplexer and a replacement for tools like tmux. you can run herdr inside tmux if you want, but we can’t support tmux-specific edge cases or weaken herdr’s nested-launch guard to accommodate them. closing this as not planned.

PR #2137, opened 2026-07-31
fix: ignore leaked herdr env inside tmux for nested-launch guard

WSHAPER:
## What

The nested-launch guard blocks `herdr` when `HERDR_ENV=1` is set. That marker is injected into every herdr-managed pane, but it leaks into tmux's server-global environment when a tmux server is started from inside a herdr pane. Every session on that server then inherits `HERDR_ENV`, so the guard false-positives and refuses to launch herdr in unrelated tmux sessions that are not herdr panes. See #2135.

## Change

herdr panes never set `TMUX`, so a set `TMUX` means the `HERDR_ENV` marker was inherited from the tmux server rather than set by herdr for this shell. The guard now ignores the marker while `TMUX` is set.

Targeted heuristic for the documented tmux coexistence case (other multiplexers can leak similarly and are intentionally out of scope here). Trade-off: a launch that is genuinely nested *through* tmux (herdr pane -> tmux -> herdr) is no longer blocked. I think accepting that rare case is worth fixing the common false positive, but happy to drop this in favor of a deeper check (e.g. validating `HERDR_PANE_ID` against the live server) if that direction is preferred.

## Validation

- `cargo fmt --check` clean
- unit: adds `nested_herdr_does_not_block_when_env_leaked_from_tmux`; existing nested-guard unit tests pass
- integration: `auto_detect_respects_nested_guard_before_auto_attach` and `explicit_client_command_respects_nested_guard` pass; both now `env_remove("TMUX")` so they pin the non-tmux nested case deterministically

`just`/`cargo-nextest` were unavailable in this environment, so `just check` was not run; I used `cargo test` for the `bin`, `cli`, and `auto_detect` targets. 14 pre-existing failures in `detect::manifest*`, `server::headless`, and `workspace` reproduce identically on clean `upstream/master` (agent-detection manifest tests needing network/fixtures) and are unrelated to this change.

refs #2135

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,10 @@ use crossterm::execute;
 
 pub(crate) const HERDR_ENV_VAR: &str = "HERDR_ENV";
 pub(crate) const HERDR_ENV_VALUE: &str = "1";
+// `TMUX` is set inside every tmux pane. herdr panes never set it, so a set
+// `TMUX` means `HERDR_ENV` was inherited from a tmux server whose launch shell
+// was a herdr pane, not set by herdr for this shell.
+const TMUX_ENV_VAR: &str = "TMUX";
 const NESTED_HERDR_MESSAGES: [&str; 6] = [
     "inception detected. we need to go deeper... said no one ever.",
     "recursion is a pathway to many abilities some consider to be... unnatural.",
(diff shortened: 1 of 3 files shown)


ogulcancelik:
thanks for the contribution. herdr is a terminal multiplexer and a replacement for tools like tmux. you can run herdr inside tmux if you want, but we can’t support tmux-specific edge cases or weaken herdr’s nested-launch guard to accommodate them. closing this as not planned.

Issue #2154, opened 2026-08-01
probe_stream_closed: Ok(_) arm returns Ok(true) (closed) instead of Ok(false) (open), silently drops the read byte


Sertug17:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

In `src/ipc.rs`, `probe_stream_closed` matches on `stream.read()`. The `Ok(_)` arm a successful byte read, meaning the stream is alive returns `Ok(true)`, the same value as the `Ok(0)` EOF arm:

```rust
Ok(0) => Ok(true),  // EOF peer closed
Ok(_) => Ok(true),  // live read treated as closed; byte silently dropped
```

This causes `local_stream_peer_closed` → `should_stop_connection` to report the connection closed whenever a byte arrives during the nonblocking probe, and the byte is lost.

### Expected behavior

`Ok(_)` should return `Ok(false)` a successful read means the stream is open, not closed.

```rust
Ok(_) => Ok(false),
```

### Reproduction

1. Open `src/ipc.rs`, function `probe_stream_closed` (~line 158).
2. Observe: `Ok(0) => Ok(true)` and `Ok(_) => Ok(true)` both arms return the same value.
3. Any byte arriving during the nonblocking probe triggers early connection teardown via `should_stop_connection` (api/server.rs ~line 686) and drops the byte.

### Impact

If a byte arrives on the subscription stream during the probe, the server tears down the connection prematurely and silently discards the data. The logic conflates "byte received" with "connection closed", which are opposites.

### Environment

- **Herdr version:** HEAD (master branch, source inspection)
- **Update channel:** git / source
- **OS:** All Unix platforms (bug is in `#[cfg(unix)]` code path)
- **Terminal:** N/A — logic error confirmed by static code analysis, not runtime behavior

PR #2155, opened 2026-08-01
fix(ipc): probe_stream_closed returns Ok(false) for successful reads

Sertug17:
`Ok(_)` arm in `probe_stream_closed` returned `Ok(true)` (stream closed) instead of `Ok(false)` (stream open). A successful read means the peer is alive; treating it as EOF caused premature connection teardown and silent byte loss.

refs #2154


ogulcancelik:
thanks for all contributions but please dont open prs for super simple changes unless they are p1/p2 bugs.

PR #2159, opened 2026-08-01
fix(pi): restrict state reporting to TUI sessions

rhjoh:
## Problem

Pi RPC mode reports hasUI as true because it exposes UI APIs over JSON. When an RPC process inherits a Herdr pane environment, the managed Pi extension therefore reports a headless process as the pane agent.

Reproduced with Pi 0.83.0: the original extension sent Herdr lifecycle requests from RPC mode. With this change it sends none, while TUI reporting continues normally.

## Fix

Require ctx.mode to be tui before starting lifecycle reporting. This excludes RPC, JSON, and print sessions. Bump the managed Pi integration to version 8 so existing installations receive the updated extension.

## Tests

- just ci
- regression coverage for RPC with hasUI set to true


ogulcancelik:
ty!

PR #2162, opened 2026-08-01
fix: render OSC 4 palette overrides instead of forwarding the index

hamidi-dev:
**What's broken**

A cell carrying a palette index is emitted to the host terminal as that index, so the host resolves it against its own palette. When a program in a pane redefines an entry with OSC 4, libghostty parses and stores it — `palette_color_query_response` even answers OSC 4 queries with the new value — but the redefinition never reaches the screen.

**Repro**

```sh
printf '\033]4;18;rgb:a9/b1/d6\033\\\033[38;5;18mshould be lavender\033[0m\n'
```

In a herdr pane this renders in the host's colour for index 18; outside herdr it's lavender.

**Found via** hamidi-dev/opentab#12, reported by SimonEisenhauer — a curses TUI that themes itself with `init_color`. Every theme rendered identically inside herdr while other TUIs looked fine, because those emit truecolor SGR and take the `CellColor::Rgb` path. I've shipped a workaround on my side, but since this is the root cause I thought I'd submit a PR to fix it here too.

**Fix**

Compare the render state's active palette against `COLOR_PALETTE_DEFAULT` to find exactly the entries a program overrode, and emit those as RGB. Every other index is still forwarded, so panes that never use OSC 4 render as before and keep following the host terminal's theme.

Resolving every index would be simpler, but it would replace the host's palette for users whose terminal doesn't answer herdr's colour queries — their theme would silently become libghostty's default. Only resolving real overrides can't regress anyone.


ogulcancelik:
ty!

Issue #333, opened 2026-05-28
WSL2: clipboard history (Win+V) does not work when copying from herdr


Loongphy:
### Current behavior

In WSL2, copying text in herdr via mouse selection does not write to the Windows clipboard history. The text can be pasted with Ctrl+V, but pressing Win+V shows no entry for the copied content.

### Expected behavior

Text selected and copied in herdr should appear in the Windows clipboard history (Win+V).

### Reproduction

 1. Run herdr in WSL2
 2. Open a pane with terminal output
 3. Mouse-select a line of text
 4. Press Win+V — clipboard history is empty
 5. Press Ctrl+V — text pastes correctly

### Impact

Clipboard history is broken for WSL2 users. This prevents reusing previously copied text and breaks the expected Windows clipboard workflow.

### Environment

- Herdr: 0.6.4
- OS: Windows 11 + WSL2 (Debian)
- Shell: bash
- Terminal: Windows Terminal

### Contribution intent

- [x] I intend to open a PR for this issue.
- [x] I understand this does not mean the issue is approved.

ogulcancelik:
could you test the raw osc 52 path in the same windows terminal + wsl environment?

```bash
command -v wl-copy xclip xsel
printf '\\e]52;c;%s\\a' "$(printf 'raw-osc52-history-test' | base64 -w0)"
```

after running the second command, please check both:\n\n1. `ctrl+v` pastes `raw-osc52-history-test`\n2. `win+v` shows `raw-osc52-history-test` in clipboard history\n\ni tested this in a windows 11 vm with windows terminal + wsl. when `wl-copy`, `xclip`, and `xsel` are absent, herdr falls back to osc 52, and the copied text appears in win+v history. so i want to confirm whether your setup is using the same path or a wslg/native clipboard tool path.

sanirudh17:
When will native windows powershell be supported?

Loongphy:
> could you test the raw osc 52 path in the same windows terminal + wsl environment?
> 
> command -v wl-copy xclip xsel
> printf '\\e]52;c;%s\\a' "$(printf 'raw-osc52-history-test' | base64 -w0)"
[...]

PR #2168, opened 2026-08-01
fix: prefer native clipboard on WSL

aka-luan:
## Summary

- prefer the native platform clipboard for local WSL sessions
- keep OSC 52 preferred for SSH sessions
- retain OSC 52 as the fallback when native clipboard writing fails
- add regression coverage for local WSL clipboard routing

## Problem

Herdr 0.7.5 forced OSC 52 for every WSL session. In Hyper under WSL, the OSC 52 sequence was not applied to the host clipboard, leaving the clipboard stale after selecting text in a pane. The native WSL clipboard bridge works correctly.

## Testing

- 
- 
-  (19 passed)

### DIFF
--- a/src/selection.rs
+++ b/src/selection.rs
@@ -304,9 +304,12 @@ fn is_wsl() -> bool {
 fn should_prefer_osc52_for_env(
     ssh_connection: Option<&OsStr>,
     ssh_tty: Option<&OsStr>,
-    wsl: bool,
+    _wsl: bool,
 ) -> bool {
-    ssh_connection.is_some() || ssh_tty.is_some() || wsl
+    // WSL has a native clipboard bridge available to local terminal apps.
+    // Keep OSC 52 as the preferred path for SSH, where the remote process
+    // cannot access the attaching terminal's native clipboard directly.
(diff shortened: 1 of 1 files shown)


ogulcancelik:
thanks for the report. osc 52 is the widely supported terminal-native clipboard path and is currently the most robust option for wsl because the attaching terminal writes directly to the windows clipboard. we intentionally made it the wsl default in [#333](https://github.com/herdrdev/herdr/issues/333) after wslg tools such as `wl-copy` succeeded without populating windows clipboard history.

this appears to be a hyper limitation. hyper’s [osc 52 support request](https://github.com/vercel/hyper/issues/6153) remains open, although xterm.js added support in [xtermjs/xterm.js#4220](https://github.com/xtermjs/xterm.js/pull/4220).

this pr is therefore not a safe general fix. it changes clipboard routing for every local wsl session and would restore the behavior that caused #333. the osc 52 fallback does not prevent that regression because a wslg clipboard command can return success without updating windows clipboard history. hyper should add osc 52 support rather than changing herdr’s wsl default.

PR #2183, opened 2026-08-02
fix(cli): accept --flag=value and reordered options in pane read/wait-output

KyleCo76:
## Problem

`herdr pane read --help` and `herdr pane wait-output --help` (clap-generated)
advertise `herdr pane read [OPTIONS] <PANE_ID>`, but the hand-rolled parsers
only accept space-separated options *after* the pane id:

    $ herdr pane read w1:p1 --source=recent --lines=5
    unknown option: --source=recent            (exit 2)

    $ herdr pane read --source recent --lines 5 w1:p1
    unknown option: recent                     (exit 2; "--source" was consumed as the pane id)

    $ herdr pane wait-output w1:p1 --match ready --timeout=5000
    unknown option: --timeout=5000             (exit 2)

Failures are loud (exit 2), but the error can point at the wrong token, and
scripts written against `--help` or the usual `--flag=value` convention break.

## Fix

- `expand_equals_args` (shared helper) splits `--flag=value` into two tokens
  for value-taking options only; boolean and unknown options keep their
  attached value and still reach the unknown-option branch.
- Both parsers extracted into pure `parse_*_args` functions (same pattern as
  `parse_pane_split_args`) that accept the pane id positional in any order and
  return the schema params directly.

Unlike the worktree `--json` compatibility no-op removed from public help in
#2174, these pane options already populate supported API fields and work in
the existing pane-id-first form; only the hand parser disagrees with the
generated syntax.

Previously valid invocations keep the same defaults and wire requests.
Invalid source/format/numeric values now print the underlying parser message
directly and return CLI-usage exit 2 instead of bubbling an `io::Error`
through `main` (exit 1). A second positional now reports
`unexpected argument: X` instead of `unknown option: X`. Server and protocol
are untouched; no docs changes needed.

## Tests

- Unit tests cover both value spellings, pane-id order, attached values on
  boolean flags, and representative success/error cases; `cargo nextest run
  --locked` is green.
- CLI-level test: one invalid value per command exits 2 with the plain parser
  message instead of the old `Error: Custom { ... }` wrapper.
- Verified live against a dev server: every rejected form above now returns
  the expected output/match.

Hit this running herdr across two machines with claude/codex/kimi agent
fleets, scripting output gates on `pane read` / `wait-output`.


KyleCo76:
Changes done for pane read — its generated help advertises both detection and --raw (read_source_option(true) + flag("raw") in spec.rs), so the fallback usage now lists both. Left wait-output unchanged: its help deliberately omits detection (read_source_option(false)), and the usage string mirrors the generated help.

ogulcancelik:
ty

Issue #953, opened 2026-07-03
Page Up/Page Down doesn't work in standard terminal mode


apaz-cli:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

Currently it is impossible to forward page up/page down to a terminal in main screen mode. Herdr intercepts it to control its own scrollback buffer. Notably, programs using the alt screen (vim, nano, opencode, etc) do not have this problem.

See the code that intercepts:
https://github.com/ogulcancelik/herdr/blob/master/src/app/input/terminal.rs#L91-L132

### Expected behavior

Pressing page up/down in less/git diff should do something, like it does in a normal terminal.

### Reproduction

Open herdr in a dirty git repo and type `git diff HEAD`, then try to navigate it.

### Impact

If I want to review a diff, I have to open another terminal outside of herdr.

### Environment

- Herdr version: v0.7.1
- Update channel (stable or preview): stable
- Operating system: Debian
- Terminal: Tested multiple terminals
- Shell, if relevant: n/a
- Relevant config, if any: n/a

PR #2191, opened 2026-08-02
fix: page keys scroll pane scrollback at zsh and REPL prompts

HackAttack:
PageUp/PageDown were forwarded to the pane whenever DECCKM (application cursor) was on, assuming only primary-screen pagers enable it, but zsh's line editor also enables DECCKM, as do REPLs such as python3. The result was PageUp scrolling shell history instead of Herdr scrollback.

Bracketed paste discriminates the two: it means the app accepts typed or pasted text at a prompt, so line editors enable it and pagers do not.

Related: #953


Toby1009 (approved):

ogulcancelik:
ty!

PR #2202, opened 2026-08-02
fix(worktree): enforce exact repository identity

chsong1:
## Bug
Worktree socket mutations can act on a checkout selected only by path/branch after its repository identity has changed. An exact-permit client can also reach an older server that ignores new permit fields unless the protocol boundary changes.

## Fix
- accept exact repo common-dir, checkout path, branch, and HEAD OID permits for worktree create/remove
- resolve and revalidate Git identity immediately before mutation, including TOCTOU-resistant create/update-ref ordering
- return receipts binding the executed mutation to the observed identity; roll create back on mismatch and preserve removal failures
- bump protocol to 20 so exact-permit clients cannot mutate through older servers
- preserve legacy request behavior for protocol-20 callers that omit exact permits

## Verification
- `cargo fmt --all -- --check`
- generated protocol schema freshness and protocol-guard tests
- `cargo test worktree` (133 pass)
- `cargo test permit_tests` (3 pass)
- full `cargo test` (2911 pass before protocol-only follow-up)
- `cargo build --release`
- independent review of identity mismatch, rollback, branch-leak, and server compatibility paths

### DIFF
--- a/docs/next/api/herdr-api.schema.json
+++ b/docs/next/api/herdr-api.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "protocol": 19,
+  "protocol": 20,
   "schema_version": 1,
   "schemas": {
     "error_response": {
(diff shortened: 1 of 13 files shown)


ogulcancelik:
this is a too large and risky prs. please open an issue first.

Issue #1630, opened 2026-07-20
Claude integration always report "idle" state despite working


AnsonH:
> [!NOTE]
> TL;DR: Caused by `CLAUDE_CODE_DISABLE_TERMINAL_TITLE=1` environment variable. Unset it and it'll work

### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

Despite having Claude integration installed with `herdr integration install claude`, the status is always stuck at "✅ idle" despite it's working.

### Expected behavior

It should show "working/blocked/done" states correctly

### Reproduction

https://github.com/user-attachments/assets/06016239-1e7f-4173-9ffc-956ecdf73cd0

AI prompt: 

> Please create and run a Bash script that sleeps for 10 seconds and then echo "Hi", then wait for the echo output and reply "Hi Bash"

Followed the [Debug integration state](https://herdr.dev/docs/integrations/#debug-integration-state) section:

1. `herdr agent list | jq` -> To find the pane ID `w1:p5`
2. Run the Claude prompt. While it's working, run `herdr agent explain w1:p5 --verbose`. Somehow the status is stuck at `idle`

```
agent: claude
state: idle
manifest: remote:/Users/anson.heung/.local/state/herdr/agent-detection/remote/claude.toml 2026.07.13.1
rule: live_prompt_box (region=prompt_box_body priority=950)
evidence: "❯\n"
visible: idle=true blocker=false working=false
cached_remote_version: 2026.07.13.1
local_override_shadowing_remote: false
remote_update_status: current
evaluated_rules:
  ✗ osc_title_working priority=1100 region=osc_title state=working
    matchers: contains=Array [] regex=Array [String("^[\\x{2800}-\\x{28FF}] ")] line_regex=Array [] all=0 any=0 not=0
    region: bytes=0 preview=""
  ✗ btw_overlay_working priority=975 region=bottom_non_empty_lines(5) state=working
    matchers: contains=Array [] regex=Array [] line_regex=Array [String("^\\s*/btw(?:\\s|$)"), String("(?i)esc to close\\s*$")] all=0 any=0 not=0
[...]

PR #2205, opened 2026-08-02
docs: add Claude Code state detection requirement 

AnsonH:
Clarifies in the `Integrations > Claude Code` section that `CLAUDE_CODE_DISABLE_TERMINAL_TITLE=1` should NOT be set or it'll break the state detection.

refs #1630

### DIFF
--- a/docs/next/website/src/content/docs/integrations.mdx
+++ b/docs/next/website/src/content/docs/integrations.mdx
@@ -53,10 +53,10 @@ herdr integration uninstall grok
 
 Herdr uses integrations in two different ways:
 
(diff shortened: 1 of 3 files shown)


AnsonH:
@ogulcancelik Appreciate your help to re-open this PR 🙇 

Didn't use `fix:` because it's not fixing a bug in Herdr - purely a documentation change so `docs` is an appropriate conventional commit tag. Thanks so much!

ogulcancelik:
i dont think that doc is the correct place to have this warning. i need to think about this.

PR #2219, opened 2026-08-03
fix(resurrect): hermes session refs persist across restart

buihongduc132:
# Fix: Hermes session refs not persisted across herdr restart

## Problem

Hermes agents running in herdr panes were not having their session references persisted across herdr server restarts. This meant that when herdr was restarted, hermes sessions could not be resumed, breaking the resurrection feature.

## Root Cause

The herdr-agent-state plugin was installed in the global `~/.hermes/plugins/` directory, but hermes profiles have their own isolated plugin directories at `~/.hermes/profiles/<profile-name>/plugins/`. When hermes was started with `--profile <name>`, it only loaded plugins from the profile's directory, not the global one.

This meant the herdr-agent-state plugin was never loaded for profile-based hermes sessions, so session refs were never reported to herdr.

## Solution

1. Copy the herdr-agent-state plugin to each hermes profile's plugins directory
2. Ensure the plugin is enabled in each profile's config.yaml
3. Added comprehensive test script to verify 10 concurrent hermes agents across 3 workspaces persist correctly

## Changes

- `scripts/test-10-agents.sh`: Test script that creates 3 workspaces with 10 hermes agents (4+3+3 distribution), sends prompts, kills herdr, restarts it, and verifies all 10 session refs persist

## Testing

```bash
# Run the test
bash scripts/test-10-agents.sh

# Expected output:
# Started: 10 / 10
# Session refs before kill: 10
# Session refs after restart: 10
# PASS: 10/10 agents started, 3 WS + 10 panes restored
```

## Verification

Tested with:
- 10 concurrent hermes agents across 3 workspaces
- All agents using `--profile hermes-manager`
- Session refs verified before and after herdr restart
- All 10 sessions persisted correctly

## Notes

This is a deployment/configuration issue, not a code bug. The plugin itself works correctly when loaded. The fix ensures the plugin is available in all hermes profile contexts.

For users with multiple hermes profiles, they need to ensure the herdr-agent-state plugin is copied to each profile's plugins directory:

```bash
for profile in ~/.hermes/profiles/*/; do
    mkdir -p "$profile/plugins"
    cp -r ~/.hermes/plugins/herdr-agent-state "$profile/plugins/"
done
```

### DIFF
--- a/scripts/test-resurrect.sh
+++ b/scripts/test-resurrect.sh
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+# Integration test for herdr resurrect fix
+# Tests 2 workspaces with pi+hermes agents in isolated session
+set -euo pipefail
+
+HERDR_FIX="${HERDR_FIX:-/home/bhd/.local/bin/herdr-fix}"
+TEST_SESSION="herdr-resurrect-test-$$"
+WS1="/tmp/herdr-resurrect-ws1-$$"
+WS2="/tmp/herdr-resurrect-ws2-$$"
+LOG="/tmp/herdr-resurrect-test-$$.log"
+
+cleanup() {
+    local code=$?
+    echo "=== Cleanup ===" >&2
+    pkill -f "$TEST_SESSION" 2>/dev/null || true
+    rm -rf "$WS1" "$WS2" 2>/dev/null || true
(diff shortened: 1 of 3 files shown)


ogulcancelik:
please write the pr descriptions yourself or at least edit so i can understand what's going on here.

PR #2256, opened 2026-08-04
fix: render halfwidth katakana voiced marks

kazunari-kamata:
## Summary

Fix rendering for halfwidth katakana graphemes that use halfwidth voiced or semi-voiced marks, such as `ｶﾞｷﾞｸﾞｹﾞｺﾞ` and `ﾊﾟﾋﾟﾌﾟﾍﾟﾎﾟ`.

## Root Cause

Ghostty reports halfwidth katakana plus U+FF9E/U+FF9F as a wide cell with a spacer tail. Herdr previously normalized the grapheme through `unicode-width`, which reports these strings as width 1, so the pane renderer could replace them with blanks.

After preserving the grapheme, the ANSI blitter still treated it as width 1 and could draw the tail cell afterward, overwriting the voiced kana on the host terminal.

## Changes

- Preserve halfwidth katakana + halfwidth voiced/semi-voiced mark graphemes when Ghostty reports them as wide cells.
- Keep Ghostty spacer-tail cells as empty symbols instead of converting them to spaces.
- Treat these graphemes as width 2 in the terminal ANSI blitter so full and diff redraws skip covered tail cells correctly.
- Add regression tests for pane rendering, retained dirty patches, and ANSI full/diff redraw behavior.

## Validation

- `cargo fmt --check`
- `cargo test halfwidth_katakana_voiced -- --nocapture`
- `cargo test halfwidth_voiced_kana -- --nocapture`

### DIFF
--- a/src/pane/terminal.rs
+++ b/src/pane/terminal.rs
@@ -2654,8 +2654,24 @@ pub(super) fn ghostty_normalize_buffer_symbol(
     if wide == crate::ghostty::CellWide::Narrow && actual_width == 2 {
         return symbol.to_string();
     }
+    if wide == crate::ghostty::CellWide::Wide && is_halfwidth_katakana_voiced_grapheme(symbol) {
+        return symbol.to_string();
+    }
 
     ghostty_blank_symbol_for_width(wide).to_string()
+}
+
+fn is_halfwidth_katakana_voiced_grapheme(symbol: &str) -> bool {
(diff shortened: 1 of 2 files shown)


kazunari-kamata:
Replaced by #2257 to use a branch name without the agent/ prefix and a scoped commit/PR title matching the project convention.

PR #2272, opened 2026-08-04
Feat: Health Check, Pane Movement, tab/workspace/agent Colors.

TechieDheeraj:
1. Added more scrollable window for the side bar.
2. Added Color option for tabs, workspace, agent tabs
3. Added Unread option for agent tabs
4. Now the agent tab will only be read when that pane is in focus not the tab (which is the current scenario)
5. Added health check option just below the agents in the side bar which displays (battery, cpu, ram, disk for local and remote when (--remote) flag is used.
6. Panes can be moved between workspaces and tabs.
7. side bar can be slided more to the right to have more room for the agent names
8. agent tab contains:
     <workspace_name> . <tab_name> . <pane_name>
     <agent_directory_name> . <agent_name>

https://github.com/user-attachments/assets/64a4c773-8a40-45ee-80a1-dcf6db968ac4


TechieDheeraj:
@ogulcancelik , can you look at this PR ?

ogulcancelik:
@TechieDheeraj as our bot told you this is a very big pr and we don't accept feature request prs unless it's aligned on a discussion first. please open a discussion first.

PR #2273, opened 2026-08-04
fix: use distinct status shapes in spaces sidebar

### LINKED ISSUE #2260: Make sidebar status indicators colorblind-safe (distinct shapes, not just color)
### Discussed in https://github.com/herdrdev/herdr/discussions/824

<div type='discussions-op-text'>

<sup>Originally posted by **jcmuller** June 26, 2026</sup>
Right now the left spaces sidebar tells agent status apart almost entirely by color. `state_dot()` maps three different states to the same `●` glyph: blocked is red, working is yellow, idle-unseen is teal. If you are colorblind, those three collapse into "a filled dot."

The agents panel on the right does not have this problem. It calls a sibling function, `agent_icon()`, that gives every state its own shape on top of color (`◉` blocked, an animated spinner for working, `●` idle-unseen, `✓` done, `○` unknown), plus a separate text label. Same `(state, seen)` input, accessible output. The sidebar just never got the same treatment. Switching to a monochrome theme like vesper actually makes it worse, because that removes the one thing currently carrying the distinction.

A plugin cannot fix this, since plugins cannot restyle the UI. The smallest fix is in the TUI client layer: point the sidebar (and the mobile views) at the same `agent_icon` mapping the agents panel already uses, and delete the duplicate `state_dot`. Pure presentation, no server/API/protocol change. The only visible shifts are blocked `●` to `◉`, working `●` to the spinner, done `○` to `✓`, and unknown `·` to `○`; idle-unseen `●` stays put, and every glyph is one cell so the columns do not move.

I prototyped it on my fork and it passes the full suite, with new buffer-render tests asserting blocked and done render distinct glyph cells. Happy to open a PR if you are up for it. Want me to file an issue first?</div>

jcmuller:
The spaces sidebar and the mobile views told agent status apart almost only by color. Blocked, working, and done-unseen all rendered as the same filled dot, so a colorblind user saw one dot for three different states. The agents panel already solved this with agent_icon, which gives every state its own shape. Point the sidebar and mobile views at that same function and delete state_dot, the color-only mapping they used to call.

Closes #2260

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -842,7 +842,7 @@ mod tests {
         let line1 = buffer_row_text(buffer, card, card.y);
         let line2 = buffer_row_text(buffer, card, card.y + 1);
 
-        assert!(line1.starts_with(" · one"));
+        assert!(line1.starts_with(" ○ one"));
         assert!(!line1.contains("1 one"));
         assert_eq!(line2, "   main");
 
--- a/src/ui/mobile.rs
+++ b/src/ui/mobile.rs
@@ -11,7 +11,7 @@ use super::sidebar::{
     next_entry_is_indented_workspace, workspace_list_entries_expanded, AgentPanelEntry,
     WorkspaceListEntry,
 };
-use super::status::{agent_icon, state_dot};
+use super::status::agent_icon;
 use super::text::{display_width_u16, truncate_end};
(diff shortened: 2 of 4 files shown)


ogulcancelik:
i'm working on this is a standalone setting

PR #2281, opened 2026-08-04
fix(build): guide users when zig is missing

LosEcher:
## Problem

When `zig` is not installed (or not on `PATH` and `ZIG` is unset), the libghostty-vt build script panics with a bare `NotFound` error that gives no hint about what to install:

```
thread 'main' panicked at build.rs:77:10:
failed to execute zig build for vendored libghostty-vt: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

## Fix

`build.rs` now distinguishes `ErrorKind::NotFound` and reports:
- that Zig 0.15.2 is required to build the vendored `libghostty-vt`,
- how to install it (`brew install zig@0.15` on macOS, https://ziglang.org/download/ elsewhere),
- that the `ZIG` environment variable can point at the zig binary.

Other execution errors keep the previous message with the underlying error appended.

## Validation

- `cargo build` (with a working zig 0.15.2) succeeds.
- Without zig on PATH: build now fails with the descriptive message above.
- `cargo fmt --check`, `cargo clippy --all-targets --locked -- -D warnings`, maintenance script tests, and `cargo nextest run` (3030 tests) all pass locally.

Note: on this machine one plugins test initially failed due to a sandboxed home directory (`~/.config` / `~/.local/state` not writable); with `XDG_CONFIG_HOME`/`XDG_STATE_HOME` pointed at a writable cache dir the full suite passes. CI on `master` is green.


LosEcher:
The failing check is `check (macos-latest)`: it failed on `api::server::pane_graphics_stream::tests::inactive_owner_cancels_idle_stream_and_dispatches_close` with `canceled idle stream should dispatch a close: Timeout`.

This is an async timeout test in the pane graphics stream — unrelated to this build.rs change (which only alters the error message emitted when the `zig` executable cannot be found; the happy-path build command is unchanged). Locally the full suite passes (3030/3030). Could a maintainer re-run the macOS job to confirm it is a flake? I can't re-run it as an external contributor.

ogulcancelik:
@LosEcher the implementation review is clean. please rebase onto current master and rerun checks and both review bots.

JJLiebig (approved):

Pimpmuckl:
@LosEcher thank you!

PR #2282, opened 2026-08-04
feat(ui): add distinct status indicators

ogulcancelik:
## Summary

- add a dots/symbols agent-status indicator setting, with dots remaining the default
- use distinct static symbols across sidebar, navigator, and mobile status surfaces
- expose the option in Settings and document `ui.status_indicators`

## Testing

- `just check`
- `just website-build`

Refs #2260

### DIFF
--- a/docs/next/CHANGELOG.md
+++ b/docs/next/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Added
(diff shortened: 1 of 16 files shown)


abhishekvtangod:
Any reason why the default is dots? I'd prefer to keep symbols as the default.
They're more recognizable at a glance and make it easier to distinguish statuses quickly.

I raised a PR to make that change. Could you please take a look when you have a chance?
https://github.com/herdrdev/herdr/pull/2442

PR #2285, opened 2026-08-04
fix(client): re-assert host mouse modes after a terminal reconnect

dengoswei:
### Current behavior

A host terminal can be re-created underneath a live client. Web VS Code restores a
reconnected terminal from a serialized snapshot that brings back the mouse tracking
mode without the SGR encoding Herdr asked for. Every mouse report then arrives in an
encoding the client cannot parse and leaks into the focused pane as text, so moving
the mouse types garbage. Recovering requires detaching and reattaching.

### Reproduction

1. Attach to a Herdr session in the VS Code web terminal.
2. Reload the browser tab.
3. Move the mouse over the pane — escape-sequence garbage is typed into it.

### Fix

Re-assert the host mouse mode set on regained focus and on resize, both of which fire
when a client reconnects. `set_mouse_capture` already clears host mouse reporting
first, so this also drops the stale tracking mode.

### Checks

`cargo fmt --check`, `cargo clippy --all-targets --locked -- -D warnings`, and the
`x86_64-pc-windows-msvc` clippy pass are clean on this branch.

`cargo test --locked` leaves 6 failures on my machine. All 6 fail identically on an
unmodified checkout of the same base commit: five `update::tests` cases that talk to a
running Herdr server (I have one running) and
`workspace::tests::generated_workspace_ids_are_short_base32_handles`, which passes in
isolation. Nothing in `raw_input` or `client` fails.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -566,6 +566,16 @@ fn set_mouse_capture(enabled: bool) -> io::Result<()> {
     }
 }
 
+/// Re-apply the host mouse mode set after the host surface may have been
+/// re-created underneath us. `set_mouse_capture` clears host mouse reporting
+/// first, so this also drops a tracking mode that a reconnect restored behind
+/// herdr's back in an encoding the client cannot parse.
+fn refresh_host_mouse_capture(enabled: bool) {
+    if let Err(err) = set_mouse_capture(enabled) {
+        warn!(err = %err, "failed to re-assert host mouse capture");
+    }
+}
+
 fn restore_terminal_state(
     reset_modify_other_keys: bool,
     reset_host_color_scheme_reports: bool,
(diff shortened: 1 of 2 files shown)


ogulcancelik (changes requested):
`refresh_host_mouse_capture(state.mouse_capture_active)` regresses Unix direct-terminal attach when `ui.mouse_capture` is disabled. That attach mode still enables host mouse reporting for its own scrolling/input path, but the first resize reapplies `false` and turns it off. Please track/reapply the effective host mode rather than the config-derived pane-forwarding state, add a direct-attach regression test, then rebase onto current master (including #3167's client-loop change) and rerun checks and both review bots.

JJLiebig:
@akbash-bot can you check if this is still relevant after the client/server refactor + recent fixes to input handling?

If yes, please tag the OP and request them to recreate this on current master then rebase + update.

Especially given VS Code Web might move quite quickly, I want to make sure we're not re-visiting some corpses.

JJLiebig:
Superseded by #4079, rebased and rewritten for the current client/server architecture. The replacement preserves @dengoswei as the commit author and addresses the direct-attach regression raised here.

Issue #285, opened 2026-05-25
Track Zig 0.16 support


chenrui333:
## Current behavior

Herdr's source build currently targets Zig 0.15 in packaging work. Homebrew's proposed formula depends on `zig@0.15` for now.

## Desired behavior

Track what is needed for Herdr and the vendored `libghostty-vt` build to work with Zig 0.16 once distros and package managers begin moving to it.

## Why this belongs in Herdr

Having a tracking issue gives Homebrew and other package maintainers a canonical upstream reference for Zig 0.16 readiness instead of carrying that discussion downstream.

Reference: https://github.com/Homebrew/homebrew-core/pull/283413

huyz-git:
Update:

https://github.com/ghostty-org/ghostty/pull/12726
https://github.com/ghostty-org/ghostty/commit/7aa9591746ffa4d2eee458960c76554352832595
`ghostty` had merged support for Zig 0.16 on 2026-07-22

https://github.com/Uzaaft/libghostty-rs/pull/67
`libghostty-rs` has an open PR for supporting Zig 0.16, opened on 2026-07-23

Uzaaft:
uzaaft/libghostty-rs#67 got merged, and I'm doing a pass over any new API surfaces before I tag a release. Lmk if I can help out in any way, shape or form.

eejd:
Packaging herdr for MacPorts (macOS) — here's what I found investigating a Zig 0.16 re-vendor, which may help unblock this:

- ghostty merged Zig 0.16 in `7aa9591746ff` (#12726, 2026-07-22, milestone 1.4.0). ghostty `main` (currently `3790fb78fecb`) has `minimum_zig_version = "0.16.0"`.
- The vendored `libghostty-vt` in this repo is pinned at `c5a21edfcbc2d5b46540ad91b7980aca31f5f1f3` (bumped 2026-07-16, per `vendor/libghostty-vt.vendor.json`) — five days before the 0.16 merge. It's a **clean 630-commit fast-forward** to ghostty `main`, zero divergence to reconcile.
- `-Demit-lib-vt` is still present and actively built on `main` (`src/build/Config.zig`).
[...]

PR #2299, opened 2026-08-04
feat: add a command palette and move panes from it

cameronsjo:
Follows up on discussion [#2283](https://github.com/herdrdev/herdr/discussions/2283).

Opening this for visibility rather than to jump the queue. It exceeds the
automated intake budget (21 files, ~1400 lines, `feat:` title), so it will close
automatically. The branch is there if the idea is worth pursuing, and I am happy
to cut it down, split it, or drop it entirely on a maintainer's word.

## What this adds

**A command palette on `prefix+/`.** Its own overlay, listing every runnable
action alongside that action's shortcut. Typing filters; arrows move the
selection, `enter` runs it, `esc` closes, and clicking a row runs it.

Matches rank by quality (exact name, then prefix, then word start, then
substring), so typing `new tab` resolves to the command actually named that
rather than to a longer command containing it. Ordering by source position would
get that right only by accident.

**Three commands that relocate the focused pane**, to an existing space, a new
space, or a new tab. `pane move` already existed in the JSON API and the CLI,
but had no keyboard path.

Choosing an existing space arms the session navigator as a destination picker:

| Row | Result |
| --- | --- |
| `+ new space` | the pane gets a space of its own |
| a space | the pane arrives in a new tab there |
| a tab | the pane splits into that tab |

Panes are not offered as destinations. A pane is where a move lands, not
somewhere to send one, and which pane it splits beside is layout rather than a
destination. The line above the footer states what accepting the selected row
will do, since nothing else on screen distinguishes a space row from a tab row.

## Notes for review

- **The keybind reference is untouched.** Its entries gained the action they
  describe so the palette can source one inventory, which means a new keybind
  appears in both without a second list to maintain. `?` itself renders and
  behaves exactly as before.
- **`PROTOCOL_VERSION` is unchanged.** `Mode` and `NavigatorTarget` carry no
  `Serialize` derive, so neither crosses the wire, and `pane.move` is the
  existing API method.
- **The three move commands ship unbound**, reachable only from the palette. No
  default keys are claimed beyond `prefix+/` itself.
- **A move that changes nothing now reports it.** Moving a pane into the tab it
  already occupies returns a success carrying `changed: false`; the only caller
  reading these responses was looking for an `error` member and found none, so
  nothing reached the user.
- The palette's initial selection is seeded after the move is armed. Arming
  changes which rows exist, so seeding first picked an index against a list that
  then changed underneath it.

## Verification

`just check` equivalent in a Linux container, since the dev machine is on macOS
26 where libghostty-vt's pinned Zig 0.15.2 cannot link its build runner (#285):
`cargo fmt --check`, `cargo clippy --all-targets --locked -D warnings`, and
`cargo nextest run` under the macOS CI filter (`not binary(live_handoff)`).
**3177 passing**.

All nine `scripts/test_*` maintenance modules pass, split across host and
container by tooling (two need `cargo`, two need `node`; no single environment
has both).

Docs updated under `docs/next/` in en/ja/zh-cn, plus two `docs/next/CHANGELOG.md`
entries and the `config-reference.json` keybinding entry.


cameronsjo:
@ogulcancelik - Opened a discussion, and I'm kind of sort of just over the line. I know the bot gave me a smack down, but would appreciate some human eyes. Thanks for the consideration. I'd really not like to maintain my own fork, and the plugin architecture wouldn't support this fully. 😞 

Not so ninja edit: the big win here is that it really just gives the user an easy way to not have to use the cli commands to do this, and makes it so that I don't have to bloat my agents' context with me throwing darts at an unfamiliar dartboard when we're working on something else.

ogulcancelik:
@cameronsjo bot did the correct thing here, since this is a big feature request. not gigantic but still not a simple bug fix. discussions exist for me to triage, so let's see how much attention your idea gets. if it gets enough upvotes and comments i'll consider it.

cameronsjo:
@ogulcancelik Agreed, not saying the bot did the wrong thing. And I expected it based on the AGENTS.md and CONTRIBUTING.md. 

Appreciate the human eyes regardless.

PR #2301, opened 2026-08-04
fix: add positive idle detection for kiro-cli prompt

smileynet:
## Summary

- Kiro's idle prompt is now positively detected (instead of falling back to a generic heuristic)
- `not` guards prevent false idle when working indicators are in the same screen region
- Regression tests cover idle, working, overlap, stale scrollback, and blocked precedence

## Root cause

Currently there is no positive confirmation for idle state in the kiro manifest, only a fallback rule that is susceptible to false negatives due to text in scrollback.

## Evidence

Captured via `herdr agent read <pane> --source detection --format text` against kiro-cli 2.13.x on Windows. Verified with `herdr agent explain <pane> --json`.

State transitions confirmed through the local-override hot-reload loop: idle → working → blocked → idle. No false positives from stale scrollback, short terminals, or narrow panes.

Idle:

\\\	ext
 ask a question or describe a task ↵
                                      /copy to clipboard
\\\

Working:

\\\	ext
 Kiro is working · Type to steer · Ctrl+S to queue
\\\

## Design

- Priority 200 overrides stale working text in scrollback.
- Region 5 handles line wrapping in narrow panes.
- `not` guards prevent false idle when positive working indicators are in the same region.

## Validation

- `cargo test --bin herdr kiro` (9 passed)
- Live pane verification via the local-override hot-reload loop
- State transitions confirmed: idle → working → blocked → idle

## Notes

- `website/agent-detection/kiro.toml` synced with `src/detect/manifests/kiro.toml` in this PR.
- Discussion #982 proposed matching the `▸` prompt character via regex. This rule matches footer text instead to avoid false positives from `▸` appearing in output.

refs discussion #982


ogulcancelik:
@smileynet could you remove tests? i'm slowly trying to remove integration tests as they do not have meaningful purpose for fast changing patterns

smileynet:
Done, removed the manifest tests. 😄

ogulcancelik:
ty

Issue #1537, opened 2026-07-17
herdr ui inaccessible using mouse or touch


patrickwilliams2345:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

after a period of time the ui becomes unresponsive to mouse or touch. the ui still responds to keyboard input. so i can use the keyboard to navigate the ui.

### Expected behavior

ui is always responsive to mouse or touch.

### Reproduction

start herdr and after a period of idle time the ui becomes unresponsive to mouse or touch input.

### Impact

i have to restart herdr all the time to get the ui back with mouse or touch input.

### Environment

- Herdr version:0.7.4
- Update channel (stable or preview):stable
- Operating system:macos tahoe
- Terminal: a bunch of different ones
- Shell, if relevant:zsh
- Relevant config, if any:

zhang17-24:
This matches the same class of problem that PR #2311 fixes. I reproduced it and recorded the details in `herdr-mouse-garbage-investigation.md` and `herdr-mouse-garbage-reproduction.md`.

**Reproduction** (verified on herdr 0.8.0, macOS arm64, VS Code 1.131 integrated terminal):
1. Put the host terminal in DEFAULT mouse encoding: `printf '\033[?1006l' > $(tty)` (xterm.js reports X10, `ESC[M`, whenever `?1006h` is off).
2. Move the mouse over herdr. The focused pane input fills with garbage like `CN1CQ2CS3...` (button `C` = 35+32, then two coordinate bytes) and the mouse is unusable.

**Root cause** (from the investigation): herdr's Unix raw-input parser only handles SGR mouse reports (`ESC[<...`); it has no X10 (`ESC[M`) branch, so X10 reports are split into `Unsupported` + text keys typed into the pane. In addition, herdr never re-asserts `?1006h` once enabled (capture-state cache), so once the host drops to DEFAULT encoding it stays there and keeps emitting garbage.

**PR #2311** fixes both: it parses X10 reports as mouse events, and on detecting X10 forces the host back to SGR (`?1006h`, debounced). Verified end-to-end: no more garbage, and the host recovers to SGR automatically.

PR #2311, opened 2026-08-04
fix(input): parse x10 mouse reports and re-assert host sgr

zhang17-24:
## Summary
- Parse X10 (`ESC[M`) mouse reports on Unix so they are treated as mouse input instead of being typed into the focused pane as garbage.
- When the client detects X10 (the host terminal dropped to DEFAULT mouse encoding), force the host back to SGR (`?1006h`), debounced, so the encoding drop cannot persist.

## Reproduction
1. Run herdr in a host terminal that reports X10 (`ESC[M`), e.g. VS Code integrated terminal after the mouse encoding drops to DEFAULT (SGR off: `printf '\033[?1006l' > $(tty)` from the host, or after a terminal reset).
2. Move the mouse. Before: the focused pane input filled with garbage like `CN1CQ2CS3...`. After: reports parse as mouse events.

## Test plan
- Unit tests for X10 parsing, incomplete-report handling, detection, and the SGR re-assert debounce.
- Verified end-to-end in VS Code: X10 reports parse as `Mouse` (no garbage) and `?1006h` is re-asserted automatically.

refs #1537

### DIFF
--- a/src/client/input.rs
+++ b/src/client/input.rs
@@ -189,7 +189,9 @@ fn idle_flush_timeout_ms(
     host_mouse_capture_active: bool,
 ) -> i32 {
     if host_mouse_capture_active
-        && (framer.has_pending_lone_escape() || framer.has_pending_incomplete_sgr_mouse_sequence())
+        && (framer.has_pending_lone_escape()
+            || framer.has_pending_incomplete_sgr_mouse_sequence()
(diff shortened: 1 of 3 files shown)


ogulcancelik:
superseeded by https://github.com/herdrdev/herdr/pull/2312

PR #2322, opened 2026-08-04
fix(input): allow escape as a prefix-mode keybinding

fraidev:
Hey!

Esc always cancels prefix mode before keybinds ran, so copy_mode = "prefix+esc" (like my tmux bind Escape copy-mode) never worked even though the config accepted it.

This treats Esc as a normal prefix RHS when bound, and unbound Esc still cancels prefix mode.

Happy to adjust anything if needed.


fraidev:
Hey @ogulcancelik, could you take a look at this? I’m using a fork for now because I’m really used to this keybind in tmux.

fraidev:
Gentle reminder @ogulcancelik

ogulcancelik:
hey sorry for waiting long on this, but this feels weird. if you map esc as prefix then you can't leave out of prefix mode. i don't think we also had many requests to bind prefix, if we do i'd rethink how this could be done but right now i need to defer this, sorry and thanks for spending time on it

fraidev:
@ogulcancelik No problem, that is fair. I am thinking a lot about using a new keybind for this. 
Btw, can you please look at this one as well? https://github.com/herdrdev/herdr/pull/3057 (prefix+tab and prefix+shift+tab stopped working for me in WezTerm)
I didn't know about the new rules about contributing.

fraidev:
> if you map esc as prefix then you can't leave out of prefix mode.
The behavior of tmux is allow it if is a keybind.

For example, if `prefix + Esc` is bound to enter copy mode:

* `prefix + Esc` → execute the `copy-mode` binding
* `prefix + something_that_is_not_a_keybind + Esc` → cancel prefix mode

So `Esc` only cancels prefix mode when there isn't a matching keybind for the current sequence.

This PR should provide it for herdr, but I totally understand not maintaining that weird case.

fraidev:
Hi again, @ogulcancelik, can you take a look at this edge case again?

I still using a fork of 0.8.2 with it because I really like this keybind.

PR #2330, opened 2026-08-04
fix: fence terminal input under controller ownership

rand:
## Summary

- add a server-wide per-terminal controller fence when a writable terminal attachment acquires control
- reject ordinary TUI, pane API, agent, paste, mouse, and delayed-input paths while that fence is active
- admit input only from the matching controller and release ownership with compare-and-clear disconnect semantics
- add an opt-out for controller takeover while preserving the current takeover-compatible default

## Why

`terminal session control` currently owns the direct-attach route, but ordinary input routes can still write to the same PTY. That means a purported sole writer is not actually exclusive. This patch makes controller ownership authoritative at the terminal runtime boundary.

## Validation

- `just test-one terminal_control`: 15 passed
- `just test-one terminal_attach`: 13 passed
- `cargo fmt --check`: passed
- `cargo clippy --all-targets --locked -- -D warnings`: passed
- full `just ci`: blocked by pre-existing macOS failure in `live_handoff_keeps_unmanaged_agent_name_bound_to_saved_session`; the same test fails identically at untouched base `93e0086f` with `agent_not_found` for `w1:p1`

Opened as draft because the repository-wide gate is not green on the unmodified base. The controller-specific suites are green.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -142,6 +142,7 @@ pub struct App {
     pub(crate) full_redraw_pending: bool,
     pub(crate) overlay_panes: HashMap<crate::layout::PaneId, OverlayPaneState>,
     pub(crate) local_terminal_notifications: bool,
+    pub(crate) allow_terminal_control_takeover: bool,
     /// Whether this process applies `AppEvent::PrefixInputSource` to the host input source.
     /// The headless server sets this to false: the switch belongs to the foreground client,
     /// even when an App-internal drain consumes the event before the forwarding drain.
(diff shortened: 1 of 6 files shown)


ogulcancelik:
this is not a bug fix, it's a feature request. open a discussion first

rand:
Thanks for the correction. I opened the feature discussion at https://github.com/herdrdev/herdr/discussions/2332 and am treating this implementation as superseded pending agreement on semantics. The discussion also incorporates the atomic authorization/enqueue issue found in review.

PR #2335, opened 2026-08-05
fix(persist): fall back to a direct write when session.json rename crosses a file system

vidhanio:


ogulcancelik:
this feels like a very niche weird edge case that become a 220 line of round fixes, i don't see why you would only mount the session-file but not the whole dir, so sorry but i'm going to close it.

vidhanio:
I use impermanence with nix (wipe on reboot) and would prefer to only persist what is absolutely required, excluding other files in the herdr config. I have just realized that impermanence has a mount `method = "symlink"` which seems to be natively supported by herdr so this isn't actually needed to be fixed in herdr, sorry about that.

Issue #1431, opened 2026-07-15
Ctrl+[ does not close workspace navigate mode


haoxianhan:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

After pressing `prefix+w` to enter workspace navigate mode, `Ctrl+[` does not close navigate mode.
Pressing `Esc` closes it normally.

### Expected behavior

`Ctrl+[` should behave like `Esc` and return to the active pane.

### Reproduction

1. Start Herdr.
2. Press `prefix+w`.
3. Press `Ctrl+[`.
4. Observe that workspace navigate mode remains open.
5. Press `Esc` and observe that it closes.

### Impact

Users who use `Ctrl+[` as the terminal equivalent of `Esc` cannot use it to leave workspace navigate
mode.

### Environment

- Herdr version: herdr 0.7.3
- Update channel (stable or preview): stable via llm-agent.nix
- Operating system:  nixos 26.05
- Terminal:  foot
- Shell, if relevant: zsh
- Relevant config, if any: none

kovalov:
@ogulcancelik 
I've got a fix ready for this: under the kitty keyboard protocol (foot), Ctrl+[ arrives as Char('[')+CONTROL.
Could do PR?

haoxianhan:
I also have a minimal fix ready, with an end-to-end regression test covering the Kitty CSI-u input sequence: https://github.com/haoxianhan/herdr/commit/9b69219.

PR #2340, opened 2026-08-05
fix(input): leave navigate mode on ctrl+[

qaz74107410:
Ctrl+[ is the terminal-level equivalent of Esc, but navigate mode only checked `KeyCode::Esc`, so the mode stayed open.

On a legacy terminal Ctrl+[ sends 0x1b and already arrives as `KeyCode::Esc`, which is why this works for some people. Under the kitty keyboard protocol, which Herdr negotiates, the modified key is reported on its own and reaches the handler as `Char('[')` with CONTROL, so the Esc check missed it. The report came from foot, which speaks that protocol.

Adds `is_navigate_cancel_key` and uses it where navigate mode tested for Esc. Ctrl+Shift+[ keeps its own meaning because the match requires CONTROL alone.

Prefix mode needs no matching change. It already falls through to `leave_command_mode` for any unbound key, so Ctrl+[ cancels it today, and adding the same check there would shadow a custom Ctrl+[ binding under Prefix dispatch.

No doc change looks needed: the keyboard docs do not currently mention Esc for navigate mode. Happy to add one if you want the alias documented.

Checks on 1.96.1: `cargo fmt --check` and `cargo clippy --all-targets --locked -- -D warnings` are clean. `cargo nextest run --locked` is 3179/3181; the two failures, `live_handoff::live_server_holds_one_pty_master_fd_per_pane` and `terminal::state::metadata::tests::metadata_clear_only_without_ttl_does_not_extend_old_ttl`, fail the same way on unmodified master in my environment. I ran this in a Linux container because zig 0.15.2 cannot build the vendored libghostty-vt on macOS 26, so `just windows-lint` is untested here; the change is not platform-gated.

refs #1431


ogulcancelik (changes requested):
requesting changes because this currently overrides valid user configuration.

`ctrl+[` is checked before navigate-mode keybinding dispatch, so any existing `ctrl+[` binding is ignored while navigate mode is open. please treat it as a fallback cancel key only when no configured binding matched.

please cover both paths: unbound `ctrl+[` cancels navigate mode, and a configured `ctrl+[` binding still runs.

qaz74107410:
Good catch, `ctrl+[` was shadowing the binding. Fixed in 6474153.

`ctrl+[` is now checked last, after workspace up/down, reserved keys, non-indexed actions, custom commands, and indexed actions, so it only cancels when nothing else matched. Esc is back to the bare `KeyCode::Esc` check it had before this PR, so its behavior is unchanged. The helper is renamed to `is_ctrl_bracket_key` and its doc comment states the fallback-only contract.

Both paths are covered:

- `app_navigate_mode_ctrl_bracket_leaves_like_esc`, unbound `ctrl+[` cancels navigate mode, and Esc still does too.
- `app_navigate_mode_configured_ctrl_bracket_binding_wins_over_cancel`, with `navigate_workspace_down = "ctrl+["` the selection moves and navigate mode stays open.

The second test fails if the check is moved back above dispatch, so it pins the ordering rather than just passing.

ogulcancelik:
@qaz74107410 the revised ordering now preserves configured ctrl+[ bindings, so my original blocker is resolved. please rebase onto current master, add the user-facing entry to `docs/next/CHANGELOG.md`, and rerun checks and both review bots; i’ll approve the updated head.

qaz74107410:
Thanks for the review.

Rebased onto current master and added the changelog entry under Unreleased / Fixed in `docs/next/CHANGELOG.md`, crediting the reporter:

    - Navigate mode now closes on `Ctrl+[` like Esc under the kitty keyboard protocol, while a configured `ctrl+[` binding keeps priority. (#1431, thanks @haoxianhan)

The rebase only touched `src/app/input/navigate.rs`, where the helper now sits next to the new tab move functions. Behavior is unchanged from the last pass: `is_ctrl_bracket_key` is still checked last, after navigate mode dispatch, and Esc keeps its bare `KeyCode::Esc` check.

Checks and both review bots have rerun on the new head and are green. Please let me know if anything else would help.

krehwell:
first time using herdr and I noticed this issue right away (using alacritty - herdr `0.8.2`)

@qaz74107410 could you rebase once more?  

@ogulcancelik the PR is ready, ci and bots are green.  may we can get this to next release

qaz74107410:
Rebased onto current master. Only the changelog conflicted this time, the fix itself replayed unchanged, and checks are rerunning on the new head.

Thanks for testing it and for the report.

krehwell:
@ogulcancelik it's a small PR I assume?  should be good to merge so we can have this in next release?

JJLiebig:
@akbash-bot can you check if this is still relevant after the client/server refactor + recent fixes to input handling?

If yes, please tag the OP and request them to recreate this on current master then rebase + update.

Issue #2371, opened 2026-08-05
Copy mode: allow held-key repeat for cursor-movement keys (keep suppressing confirm/exit repeat)


ogulcancelik:
### Discussed in https://github.com/herdrdev/herdr/discussions/1342

<div type='discussions-op-text'>

<sup>Originally posted by **markie1006** July 12, 2026</sup>
### Summary

In **copy mode**, held cursor-movement keys (arrows, `h/j/k/l`, `w/b/e`, …) do **not** auto-repeat — one keypress moves the cursor exactly one step, even while the key is held down. The same keys auto-repeat correctly in a normal terminal pane. Please allow held-key repeat for copy-mode **navigation** keys, while keeping repeat suppressed for the confirm/exit keys (as today).

### Environment

- herdr **0.7.3** (stable)
- Host: Linux (WSL2)
- Terminal: **Ghostty on macOS**, connected over SSH to the host running herdr
- Ghostty implements the Kitty enhanced keyboard protocol and emits explicit key-repeat events.

### Steps to reproduce

1. From Ghostty (macOS) → SSH → the host, attach to herdr.
2. In a normal pane, hold an arrow key or Backspace → it repeats continuously. ✅ (so repeat events *are* reaching herdr)
3. Enter copy mode (`prefix + [`).
4. Hold an arrow key (or `j`/`k`, `w`, etc.).

### Expected

While the key is held, the copy-mode cursor keeps moving (and, with a selection active, the selection keeps extending) — as in tmux / vim visual mode / iTerm2 copy mode.

### Actual

The cursor moves exactly one cell per physical press. Holding the key does nothing beyond the first step, so navigating more than a few cells requires many discrete presses.

### Analysis

This looks like an intended-but-over-broad safety measure. The 0.3.2 changelog says held-key repeat was restored in terminal panes *"without letting modal close/confirm key repeats leak into the shell."* That guard is important — but the hazard is specific to **action/confirm/exit keys**, not to navigation:

- Repeating a confirm/exit key (`Enter`/`y` = yank & leave, `q`/`Esc` = quit) is dangerous: the first event leaves copy mode and the remaining repeats **leak into the shell** as input. Suppressing repeat for these is correct.
- Repeating a pure **movement** key (arrows, `h/j/k/l`, `w/b/e`, `{`/`}`) only moves the cursor / extends the selection. It never exits the mode and never leaks. Repeat here is harmless and is the expected UX.

[...]

PR #2372, opened 2026-08-05
fix: repeat copy-mode navigation keys

ogulcancelik:
## Summary
- repeat all copy-mode cursor and scroll navigation when terminals report explicit repeat events, including arrows, Page Up/Down, Home/End, Vim-style motions, and `Ctrl-B/F/U/D`
- keep confirm, exit, prefix, selection, search-opening, and other modal actions single-shot
- preserve input-source ownership with a copy-navigation-specific consumed lease

## Reproduction
macOS Ghostty reports held keys as Kitty `Repeat` events; Windows can report repeated native presses. Before this change, those paths moved once in copy mode. The patched build repeats navigation continuously without allowing action-key repeats to escape the mode.

## Tests
- `just check`
- automated macOS Ghostty → SSH validation: held Up changed copy-mode scroll offset from 0 to 53
- regression coverage for Ghostty enhanced arrows, Windows native character repeats, Ctrl navigation, both input routers, and action-key suppression

Refs #2371

### DIFF
--- a/docs/next/CHANGELOG.md
+++ b/docs/next/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Experimental pane graphics now support bounded named layers, acknowledged full-RGBA primary-layer direct file frames on audited local terminals, owned BGRA fallback, exact pixel mouse input, and placement-only resize replay.
 
 ### Fixed
+- Copy-mode cursor and scroll navigation keys now repeat when held in terminals that report explicit key-repeat events, while confirm, exit, prefix, and other actions remain single-shot.
 - Fish `Ctrl+Alt` keybindings now work in panes after legacy Alt-prefixed control bytes are decoded with both modifiers. (#2514)
 - `herdr config check` now reports unknown built-in theme names instead of silently accepting them. (#2452)
 - macOS `herdr --remote` clients now keep the accepted bridge socket blocking, preventing an immediate disconnect after the protocol handshake. (#2478, thanks @mathijshenquet)
--- a/src/app/input/copy_mode.rs
+++ b/src/app/input/copy_mode.rs
@@ -27,6 +27,17 @@ impl App {
 }
 
 impl AppState {
+    pub(crate) fn allows_copy_navigation_repeat(&self, key: &TerminalKey) -> bool {
+        self.mode == Mode::Copy
+            && self
+                .copy_mode
+                .as_ref()
+                .is_some_and(|copy_mode| copy_mode.search.prompt.is_none())
+            && !self.is_prefix_key(key)
+            && key.generated_text.is_none()
+            && is_repeatable_navigation_key(key)
+    }
+
     pub(crate) fn enter_copy_mode(&mut self, terminal_runtimes: &TerminalRuntimeRegistry) {
         let Some(ws_idx) = self.active else {
             return;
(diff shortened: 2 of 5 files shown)


ogulcancelik:
@coderabbitai review

joonhwan:
Though I'm interested in this fix , when i `cargo run` this branch, copy mode's keys(i.e : j,k,ctrl-u,ctrl-d ...) cannot be repeated(For Windows herdr test).

ogulcancelik:
@joonhwan  Thanks for catching this you were right. The original allowlist covered only non-character navigation. 7cce478e now includes `h/j/k/l`, word and paragraph motions, `Ctrl-B/F/U/D`, and the other safe copy-mode movements, with a Windows native-repeat regression test. Could you confirm, exit, prefix, selection, and search-opening actions remain single-shot?

joonhwan:
@ogulcancelik 

> @joonhwan Thanks for catching this you were right. The original allowlist covered only non-character navigation. [7cce478](https://github.com/herdrdev/herdr/commit/7cce478e4e0421b3cd4eec96434cbbd1e59765fd) now includes `h/j/k/l`, word and paragraph motions, `Ctrl-B/F/U/D`, and the other safe copy-mode movements, with a Windows native-repeat regression test. Could you confirm, exit, prefix, selection, and search-opening actions remain single-shot?

This feature now works in windows 11 / wt.exe !  The test results follow.

- repeat check 
  - "char movemetn" - `h/j/k/l` : ok 
  - "page up" - `Ctrl+Shift+b`(Ctrl-B)  : ok (`Ctrl+b` is prefix in my case)
  - "page down" - `Ctrl+Shift+f`(Ctrl-F) or `Ctrl+f`(Ctrl-f) repeat : ok  🫨
  - "half-page up/down"  - `Ctrl+d/u`  or ` Ctrl+Shift+d/u` repeat : ok 
- no repeat check  
  -  "search" - "/" or "?" : ok 
  -  "visual selection" - "v/space" : ok 
  -  "yank" - "y/enter" : ok.  [1]  
  -  "exit cursor mode" - "q/esc",   : ok.  [1]

[1] : no input from repeating keys after exiting cursor mode, which is qood 


BTW, since I've been using herdr, i've not have any chance to use "n/N" keys successfully(like.  `10 h`) . 😢

shibayu36:
Thank you for working on the copy-mode key repeat! This behavior has bothered me too, so I'm glad to see a fix in progress.

One suggestion: could the repeat decision be based on what the key press actually did, instead of a list of keys?

As analyzed in https://github.com/herdrdev/herdr/issues/2371, what this suppression needs to prevent is: a key triggers an action that leaves the mode, the key is still held, and the remaining repeats leak into the next mode as input. If so, whether a held key is safe to repeat is decided by the action it triggered, not by which key it is bound to.

Currently `is_repeatable_navigation_key` is a list of keys that are allowed to repeat, and its char list (h/j/k/l, g, G, w, b, e, …) duplicates the dispatch list in `handle_copy_mode_key`. Adding a new navigation motion later means updating both places, and missing one silently leaves only the new key non-repeating, which is hard to notice. If instead the repeat was stopped only when the press is observed to do something that should stop it — such as leaving the mode — everything else would repeat automatically, and the allowlist would not need updating when navigation motions are added.

One more reason: if copy-mode keybindings become configurable in the future (I proposed this in https://github.com/herdrdev/herdr/discussions/587), a hardcoded key list stops working — for example, binding cursor-up to ctrl+p would leave it non-repeating. Deciding from the result of the press keeps working no matter how keys are rebound.

ogulcancelik:
review found one remaining correctness gap: a Windows console record can arrive as a single `Press` carrying `repeat_count > 1`. the copy-navigation press path handles the action once, replaces the suppress lease with `RepeatCopyNavigation`, and then ignores the `RepeatPlan`, so the remaining `repeat_count - 1` movements are lost. the current Windows test sends two separate count-1 presses and does not catch this. please replay the bundled remainder for an eligible initial press and add a regression with one count-3 record, then rebase onto current master, resolve the changelog conflict, and rerun checks and both review bots.

shibayu36:
Following up on my earlier comment suggesting that repeatability be decided from what the key press actually did, rather than from a list of keys: it turns out herdr already has the machinery for that decision, so I built a PoC on my fork to verify it with working code:

https://github.com/shibayu36/herdr/pull/2

Key points:

- InputLeaseTable (src/app/input/lease.rs) reprocesses repeats only when the context at repeat time matches the context captured at press time; copy mode simply had no context (None), which is why its repeats were always suppressed
- Adding a NonTerminal variant to InputContext (TerminalInputContext on master) and letting copy mode participate is enough: movement keys keep the mode unchanged, so contexts match and repeats fire, while confirm/exit keys like Enter/y/q/Esc change the mode, so contexts differ and repeats stay suppressed as before. No repeatable-key list is needed. Tests in the PoC verify both behaviors
- The navigator, the workspaces sidebar, and the keybinds menu reported in the comments on #2371 also ride the same mechanism with one enum variant and one mapping line each; the PoC includes them

What do you think of this direction? It would leave no key list to maintain.

Pimpmuckl:
Closing as superseded by #3487. The client shell now uses context-aware repeat leases: copy navigation repeats while exit and confirm actions remain single-shot. The old explicit allowlist path no longer exists, and the focused replacement-path tests pass.

PR #2374, opened 2026-08-05
fix: cap unterminated bracketed-paste buffering

docxology:
On flush_timeout, if the buffer starts with ESC[200~ but the ESC[201~ terminator has not arrived, the framer held indefinitely with no cap — unbounded memory growth and input hang. A truncated paste, a protocol violation, or a split terminator causes all subsequent input to be held and never forwarded.

Fix: add a MAX_BRACKETED_PASTE_BUFFER (1 MB) cap. When exceeded, flush the raw bytes without bracketed-paste framing so the pane receives the content and input flow resumes.

Regression test: `oversize_bracketed_paste_without_terminator_is_flushed_raw` verifies that an oversize buffer is flushed as raw bytes and does not grow unboundedly.

`cargo fmt --check`: clean
`cargo test --locked raw_input`: 106 passed, 0 failed


docxology:

PR #2375, opened 2026-08-05
fix: release plugin in-flight counter if the command thread never completes

docxology:
plugin_commands_in_flight was incremented at runtime.rs:119 but only decremented on the PluginCommandFinished event. If the spawned thread panicked or blocking_send failed before the event was dispatched, the counter leaked. After MAX_PLUGIN_COMMANDS_IN_FLIGHT (32) leaks, all further plugin commands fail with "maximum concurrent plugin commands reached" until restart.

Fix: use an RAII guard (PluginCommandFinishedGuard) that sends the finished event on drop. If the thread panics or the event is never explicitly sent, the guard drops and the event fires, decrementing the counter.

Regression tests: `guard_sends_finished_on_drop` and `guard_does_not_double_send_after_explicit_send` verify the guard fires exactly once.

`cargo fmt --check`: clean
`cargo test --locked`: plugin tests pass


docxology:

PR #2376, opened 2026-08-05
fix: use process-unique temp names for session saves

docxology:
save_json_to_path used `target.with_extension("json.tmp")` — a fixed name not unique per write. If two processes (e.g. a herdr server and client, or two server instances) save concurrently, they clobber each other's in-progress temp file, and one can `rename()` a file the other is still writing, producing a truncated/garbled session.json that fails to parse on next load (silent total session loss).

Fix: use a process-unique temp file name incorporating PID, an atomic counter, and nanosecond timestamp: `.{file_name}.{pid}.{seq}.{nanos}.tmp`. This ensures no two writers collide on the same temp path.

Regression tests: `unique_tmp_path_is_unique_per_call` and `save_to_paths_leaves_no_tmp_files_behind` verify uniqueness and cleanup.

`cargo fmt --check`: clean
`cargo test --locked persist`: 9 passed, 0 failed

### DIFF
--- a/src/persist/io.rs
+++ b/src/persist/io.rs
@@ -1,4 +1,5 @@
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicU64, Ordering};
 
 use tracing::warn;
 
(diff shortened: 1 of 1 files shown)


docxology:

Issue #1809, opened 2026-07-24
Key alternation for previous_tab does not work


hongquan:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

I added this in the config:

```toml
[keys]
new_tab = ["prefix+c", "f2"]
next_tab = ["prefix+n", "f4"]
previous_tab = ["f3", "prefix+p"]

[ui]
prompt_new_tab_name = false
```

But when pressing "F3", Herdr does not switch to the previous tab.
The "F2", "F4" still work as expected.
By the way, although `prompt_new_tab_name = false`, Herdr still asks for tab name when I create new tab.

### Expected behavior

Pressing "F3" will make switch to previous tab, like "prefix + b".

### Reproduction

Just add config like above.

### Impact

I have to keep pressing F4 to go to next tab, next tab, making one round to go to "previous tab".

### Environment

- Herdr version: 0.7.4
- Update channel (stable or preview): stable
- Operating system: Ubuntu 26.04
- Terminal: Ghostty
- Shell, if relevant: Fish
- Relevant config, if any: Pasted above

hongquan:
More info: With Numlock On, the F3 does not work, with Nunlock Off, the F3 works!

lyh970817:
Independent reproduction on a different terminal and OS: foot on NixOS, herdr 0.8.0 stable (this report is 0.7.4 / Ghostty).

With a binding on `f3`, pressing F3 does nothing: the binding does not fire, and the bytes do not reach the pane either. With `cat -v` running inside a herdr pane, F3 prints nothing, while the same key in bare foot prints `^[OR`. F1, F2, F4, F5 and F6 all work. Shift+F3 fails the same way.

Feeding a herdr client the exact bytes the terminal emits:

- `\x1b[15;1:1~` (F5) -> binding dispatches
- `\x1b[13;1:1~` (F3) -> nothing
- `\x1bOR` (F3, legacy SS3) -> binding dispatches

Client log for the failing case:

```
dropping unsupported escape sequence sequence="\u{1b}[13;1:1~"
```

Reproduction note: synthesizing F3 as `\x1bOR` masks this, because that is not the form the terminal emits while the kitty keyboard protocol is active.

The same `\x1b[11;...~`-`\x1b[14;...~` form was reported in #818.

boognevatz:
as a workaround you can put this into ~/.config/foot/foot.ini

[...]

PR #2378, opened 2026-08-05
fix: recognize kitty CSI-u codepoints for F1 through F12

ashokDevs:
## Summary

`kitty_codepoint_to_keycode` mapped codepoints for F13-F35 (57376-57398) but had no entries for F1-F12 (57364-57375), so those silently fell through to `None`. Terminals that send function keys as full kitty CSI-u sequences (e.g. Ghostty, which enables the kitty keyboard protocol by default) had F1-F12 dropped entirely.

## Reproduction

Confirmed live against a running herdr session: built the binary, started an isolated dev session with `previous_tab`/`next_tab` bound to F3/F4, and injected the raw byte sequences a kitty-protocol terminal sends for those keys (`\x1b[57366u` for F3, `\x1b[57367u` for F4) directly into the pane. Before the fix, both did nothing. After the fix, both correctly switched tabs.

I could not reproduce the exact asymmetry in #1809 (reporter says F2/F4 worked but F3 didn't) - in my testing the gap affects F1-F12 uniformly. This fix addresses the confirmed underlying gap either way.

## Tests

- `just lint`
- `just ci` (3066 tests, 0 failures)
- Added `kitty_f1_through_f12_codepoints_are_recognized`, replacing `unknown_kitty_functional_key_remains_unsupported` which had codified the gap as expected behavior

refs #1809

### DIFF
--- a/src/input/parse.rs
+++ b/src/input/parse.rs
@@ -262,6 +262,7 @@ fn kitty_codepoint_to_keycode(codepoint: u32) -> Option<KeyCode> {
         57361 => Some(KeyCode::PrintScreen),
         57362 => Some(KeyCode::Pause),
         57363 => Some(KeyCode::Menu),
+        57364..=57375 => Some(KeyCode::F((codepoint - 57364 + 1) as u8)),
         57376..=57398 => Some(KeyCode::F((codepoint - 57376 + 13) as u8)),
         57399 => Some(KeyCode::Char('0')),
         57400 => Some(KeyCode::Char('1')),
@@ -318,13 +319,8 @@ fn kitty_codepoint_to_keycode(codepoint: u32) -> Option<KeyCode> {
         57452 => Some(KeyCode::Modifier(ModifierKeyCode::RightMeta)),
         57453 => Some(KeyCode::Modifier(ModifierKeyCode::IsoLevel3Shift)),
         57454 => Some(KeyCode::Modifier(ModifierKeyCode::IsoLevel5Shift)),
-        value if is_kitty_functional_codepoint(value) => None,
         value => char::from_u32(value).map(KeyCode::Char),
     }
-}
-
-fn is_kitty_functional_codepoint(codepoint: u32) -> bool {
-    (57358..=57454).contains(&codepoint)
 }
 
 #[allow(dead_code)] // Reserved for the upcoming raw stdin parser.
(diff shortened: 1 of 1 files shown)


lyh970817:
A question about scope, not a review: this PR maps the kitty CSI-u codepoints (57364-57375) for F1-F12. Is it also intended to cover the case in #1809 where F3 arrives as `\x1b[13;1:1~`, the parameterized CSI tilde form rather than a CSI-u codepoint?

I have not built or tested this branch, so I am not claiming either way. I only note that my reproduction of #1809 (foot, herdr 0.8.0) produces the tilde form, and I could not tell from the description whether that path is in scope.

Happy to test a build against that reproduction if it is useful.

WAY29:
hey any review about this? need this fix

ogulcancelik:
@ashokDevs this narrow CSI-u F1-F12 fix is still needed and is not superseded by #2578. the implementation review is clean. please rebase onto current master, add the user-facing entry to `docs/next/CHANGELOG.md`, and rerun checks and both review bots.

JJLiebig:
Re CodeRabbit's outside-diff finding on `src/input/parse.rs`:

Confirmed. `parse_xterm_modified_special_sequence` maps tilde codes `2,3,5,6,15,17,18,19,20,21,23,24` but not `11`–`14`, so foot's parameterized F3 form `\x1b[13;1:1~` falls through and is reported as `RawInputEvent::Unsupported`, while the bare `\x1b[13~` at the top-level match maps to `KeyCode::F(3)`.

That is a distinct encoding family (parameterized CSI-tilde) from the kitty CSI-u codepoint mapping this PR is scoped to. Per the requested narrow CSI-u F1–F12 scope, and CodeRabbit's own suggestion to add the tilde mapping separately from the CSI-u mapping, this diff stays narrow. The parameterized CSI-tilde F1–F4 mapping is recorded as a separate follow-up for #1809.

JJLiebig (approved):
Validated end to end locally.

Rebased on current master; the diff is limited to src/input/parse.rs and src/raw_input.rs. Tests pass (input::parse 40/40, raw_input 102/102) and CI is green across linux/macOS/windows + ConPTY, conventional-commits, and CodeRabbit.

Live check: ran a real Herdr Linux client under a PTY with previous_tab = "f3" / next_tab = "f4" and injected the raw sequences. \x1b[57366;1u (F3) and \x1b[57367;1u (F4) switch tabs on this branch; the same injection against the pre-fix parser does nothing. Legacy \x1bOR works in both, so the test harness and binding are sound.

The parameterized foot CSI-tilde form (\x1b[13;1:1~) is intentionally out of scope and will be handled separately; this PR stays the narrow kitty CSI-u F1–F12 fix.

JJLiebig:
@greptileai
@coderabbitai
final review pls

JJLiebig:
Re the `docs/next/CHANGELOG.md` ask: leaving this out of the PR is intentional, not an oversight.

Per `AGENTS.md`, normal feature/fix work must not edit `docs/next/CHANGELOG.md` — it keeps long-lived branches from conflicting over one shared release file, and refreshing an older PR should drop any changelog-only diff.

The changelog is also not auto-generated from commits. `scripts/changelog.py prepare` only rolls a hand-curated `## Unreleased` section into the versioned `## [x.y.z] - date` section (it errors if `Unreleased` is empty); the user-facing prose is authored during the pre-release audit from the commit inventory. This commit's subject and `refs #1809` are that source material, so the entry gets written at release preparation.

No changelog edit here; the code mapping itself has no defect.

PR #2382, opened 2026-08-05
fix(sidebar): highlight active agent in collapsed sidebar

ianks:
## Current behavior

In the collapsed sidebar, every agent row in the bottom section renders with the same muted style, so there is no way to tell which agent is active. The workspace list in the section directly above does mark its active entry, which makes the two lists look inconsistent at a glance.

## Expected behavior

The active agent should be visually distinguishable in the collapsed sidebar, the same way the active workspace already is.

## Cause

`render_sidebar_collapsed` styles the agent position unconditionally:

```rust
let position_style = Style::default().fg(p.overlay0);
```

There is no active/selected branch, unlike the collapsed workspace rows above it and `render_agent_detail` in the expanded sidebar, which both resolve and style the active entry.

## Change

Resolve the active entry with the same `app.is_active_pane(detail.ws_idx, detail.tab_idx, detail.pane_id)` call the expanded agent panel already uses, then apply the treatment the collapsed workspace rows already use: `surface_dim` row background and the position number in `text`.

No new tokens, styles, or affordances. `state_icon` returns a foreground-only `Style`, so ratatui's patch semantics leave the row background intact on the icon cell, matching how the workspace rows behave.

## Tests

Two tests in `src/ui/sidebar.rs`:

- `collapsed_sidebar_highlights_active_agent` asserts exactly one row is marked active and the rest stay muted.
- `collapsed_sidebar_does_not_highlight_agents_without_active_workspace` asserts nothing is highlighted when there is no active workspace.

Both assert on styles rather than row order, since `agent_panel_entries` ordering is not guaranteed.

### DIFF
--- a/src/ui/sidebar.rs
+++ b/src/ui/sidebar.rs
@@ -844,9 +844,22 @@ pub(super) fn render_sidebar_collapsed(app: &AppState, frame: &mut Frame, area: 
                 break;
             }
             let position = detail_idx + 1;
-            let position_style = Style::default().fg(p.overlay0);
+            let is_active = app.is_active_pane(detail.ws_idx, detail.tab_idx, detail.pane_id);
+            let position_style = if is_active {
+                Style::default().fg(p.text).bg(p.surface_dim)
+            } else {
+                Style::default().fg(p.overlay0)
+            };
             let (icon, icon_style) =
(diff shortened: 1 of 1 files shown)


ogulcancelik:
ty!

PR #2433, opened 2026-08-06
fix: suppress transient elevated Windows terminal titles

Pimpmuckl:
## Problem

While dogfooding Herdr on native Windows, I noticed pane titles briefly flashing `Administrator: ...` whenever Codex ran a nested PowerShell tool process.

Windows temporarily rewrites the inherited console title from `<current title>` to `Administrator: <current title>` and restores it when the process exits. ConPTY emits both changes as title OSC events, and Herdr currently publishes the intermediate title immediately.

In 10 observed runs, the decoration appeared 51–78 ms after the stable title and disappeared 54–84 ms later.

## Fix

On Windows, hold only the exact transition from `<current title>` to `Administrator: <current title>` for 200 ms.

- if the title restores or changes before the deadline, never publish the transient title
- if it remains, publish it normally after the delay
- keep every other title change immediate

The Windows-specific classification stays in `src/platform/windows.rs`; the shared terminal state only handles the generic settle/cancel/promote behavior.

## Scope

- raw PTY output and agent detection are unchanged
- non-Windows behavior is unchanged
- intentionally persistent `Administrator:` titles still appear after 200 ms
- no docs change: this restores stable title behavior and adds no setting or workflow

## Validation

- `just check` passes on native Windows after rebasing onto current `master`
- 18 live API samples from 31–1090 ms during a nested PowerShell run contained zero `Administrator:` titles
- a persistent candidate remained hidden at 15 ms, appeared by 320 ms, then restored normally
- local 5.6-sol xhigh reviewers completed with no accepted findings remaining


Pimpmuckl:
Closed, there's more annoyances with nested PS spawning "Administrator: " prefixes.

PR #2270, opened 2026-08-04
feat(copy-mode): add big-word motions

jplew:
## Summary
- add tmux/Vim-style `W`, `B`, and `E` copy-mode motions using whitespace-delimited runs
- preserve existing punctuation-aware lowercase `w`, `b`, and `e` behavior
- cover shifted input, row crossings, selection extension, and terminal motion semantics

## Validation
- `just check`

## Context
Implements the big-word portion of https://github.com/herdrdev/herdr/discussions/1267.

PR #2436, opened 2026-08-06
fix(input): keep shift on letters that report only an alternate codepoint

### LINKED ISSUE #2435: shift+letter keybindings do not match in WezTerm kitty mode

- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

A `prefix+shift+<letter>` keybinding does not match. Herdr runs the `prefix+<letter>` action instead. For example, `prefix+shift+r` starts resize mode instead of a config reload. Herdr does the same action for `prefix+r` and `prefix+shift+r`.

### Expected behavior

A `prefix+shift+<letter>` keybinding matches. Herdr does a different action for `prefix+<letter>` and `prefix+shift+<letter>`.

### Reproduction

1. Set `enable_kitty_keyboard = true` in the WezTerm config.
2. Start herdr.
3. Press the prefix key (`ctrl+b` by default), then `r`. Resize mode starts.
4. Press the prefix key, then `shift+r`. Resize mode starts again.

### Impact

You cannot use `prefix+shift+<letter>` keybindings. The default `reload_config` (`prefix+shift+r`) does not work. The keybinding does the `prefix+<letter>` action.

### Environment

- Herdr version: 0.8.0
- Update channel (stable or preview): stable
- Operating system: macOS 15 (Darwin 25.6.0)
- Terminal: WezTerm 20260805-104032-4b1c3c15
- Shell, if relevant: zsh
- Relevant config, if any: WezTerm `enable_kitty_keyboard = true`

farzadmf:
WezTerm kitty mode sends shift+<letter> with no SHIFT bit, only the uppercase alternate codepoint, so it matched the unshifted binding.

Fixes #2435

### DIFF
--- a/src/config/keybinds.rs
+++ b/src/config/keybinds.rs
@@ -1313,8 +1313,16 @@ fn key_parts_match_combo(
     shifted_codepoint: Option<u32>,
     combo: KeyCombo,
 ) -> bool {
-    let (actual_code, actual_modifiers) = normalize_key_combo((actual_code, actual_modifiers));
+    let (actual_code, mut actual_modifiers) = normalize_key_combo((actual_code, actual_modifiers));
     let (expected_code, expected_modifiers) = normalize_key_combo(combo);
+
+    // WezTerm kitty mode sends shifted letters with no SHIFT bit, only the
+    // uppercase alternate codepoint; restore SHIFT so shift+<letter> still matches.
+    if let KeyCode::Char(c) = actual_code {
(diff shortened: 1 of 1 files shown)


farzadmf:
Hopelessly pinging @ogulcancelik similar to #2270 as this is a big blocker for me on Wezterm

EDIT: by "similar to ...", I mean I think pinging you allowed the PR to go through, hopefully same will happen to this

Issue #962, opened 2026-07-03
`pane read --source recent` / `recent-unwrapped` return empty output on Windows while `visible` has content


inbeomheo:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

On Windows, `herdr pane read <pane> --source recent` and `--source recent-unwrapped` print nothing (exit 0) even when `--source visible` returns the pane content just fine.

Because of this, any output that has scrolled out of the viewport is unrecoverable through the API: in a test with a 40-line agent response, the top 8 lines scrolled off and neither recent source could retrieve them (both returned empty).

### Expected behavior

`--source recent` should return recent scrollback as rendered, and `--source recent-unwrapped` the unwrapped recent text, as described in the docs.

### Reproduction

1. `herdr workspace create --no-focus --label bugtest` (take the root pane id from the JSON response, e.g. w9:p1)
2. `herdr pane run w9:p1 "echo hello123"`
3. `herdr pane read w9:p1 --source visible --lines 8` → shows the prompt and hello123
4. `herdr pane read w9:p1 --source recent --lines 10` → empty, exit 0
5. `herdr pane read w9:p1 --source recent-unwrapped --lines 10` → empty, exit 0

### Impact

Reading other panes is limited to the current viewport, so long agent responses are partially lost to automation. This also breaks `wait output --source recent` matching semantics (see my separate wait report). Workaround: keep agent replies short or run agents headless.

### Environment

- Herdr version: 0.7.1-preview.2026-06-30-3459798b606d
- Update channel (stable or preview): preview
- Operating system: Windows 11 Home 10.0.26300 (Korean locale)
- Terminal: Windows Terminal
- Shell, if relevant: PowerShell 5.1 (default pane shell)
- Relevant config, if any: default config

PR #2474, opened 2026-08-07
fix: avoid redundant Windows recent-history snapshots

Pimpmuckl:
## What changed

There was a bit of a performance issue at times on Windows, especially in panes with frequently updating agent UIs which made the entire herdr control sluggish with multi-second delays and very slow input parsing. We could reproduce it and narrow it down to the recent-history fallback added for #962.

The earlier fix restored empty `recent` reads by taking a visible-screen snapshot after every PTY batch. That included cursor-only redraws where scrollback had not changed. Each snapshot refreshed render state, walked the visible rows, and allocated a new snapshot while holding the terminal lock.

In the live reproduction, this kept the Herdr server at 82.6% of one core over 8.5 seconds while scrollback grew by only six rows. Timing the path showed that the fallback scan took 123 to 166 microseconds after each PTY read, compared with 28 to 40 microseconds for Ghostty parsing and 15 to 19 microseconds for PowerShell working-directory reporting.

This change keeps the fallback, but avoids rebuilding its snapshot when Ghostty reports no scrollback movement. It records one pending refresh and only performs it if a recent-history read needs the fallback.

A tracked Ghostty row preserves the amount of history movement across pruning and reflow. This also covers the case where scrollback prunes and regrows to the same total in one batch.

Zero-scrollback panes, history changes, manual scrolling, resize, alternate-screen behavior, and fallback recovery keep their existing behavior.

## Before and after

On the same isolated debug-build workload with two runs of 1500 paced PTY writes:

Before:

- `pty.ghostty_write` averaged 419 microseconds across 5848 samples
- the slowest sample was 4988 microseconds

After:

- `pty.ghostty_write` averaged 230 microseconds across 5854 samples
- the slowest sample was 1687 microseconds

That is a 44.9% reduction in the affected PTY write path. The live CPU sample and the paired timing measure different things, so this only claims the paired result.

This is a controlled paired test, in real world use this is likely a lot better (but didn't run extra tests).

## Validation

- reproduced and covered the real `2288 -> 1145 -> 2288` prune and regrow case
- all 11 Windows recent-history fallback tests
- Linux cross-target `cargo check`
- fresh two-job `just check`
- local dual 5.6-sol-xhigh reviewers green
- isolated Windows ConPTY smoke

refs #962

### DIFF
--- a/src/ghostty/mod.rs
+++ b/src/ghostty/mod.rs
@@ -768,6 +768,9 @@ pub fn encode_focus(event: FocusEvent) -> Result<Vec<u8>, Error> {
 
 pub struct Terminal {
     raw: ffi::GhosttyTerminal,
+    max_scrollback: usize,
+    #[cfg(windows)]
+    tracked_row: ffi::GhosttyTrackedGridRef,
     callback_state: Box<TerminalCallbackState>,
     kitty_fingerprints: Mutex<HashMap<u32, KittyImageFingerprintEntry>>,
     kitty_empty_generation: Cell<Option<u64>>,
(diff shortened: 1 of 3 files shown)


ogulcancelik (approved):

ogulcancelik:
ty!

Issue #2459, opened 2026-08-07
Windows + Git Bash: idle CPU still scales with pane count (~0.11 cores/pane); PowerShell panes cost ~0.01


hernant:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

The server sits at ~3.1 of 8 logical cores with every pane idle — no output, no input, no agent activity.

Adding six idle Git Bash panes adds ~0.7 cores; closing them returns to baseline. Changing only `terminal.default_shell` to PowerShell and repeating the exact same steps adds almost nothing.

Marginal cost per idle pane, same machine, same running server:

| Pane shell | Runs | Cores per idle pane |
|---|---|---|
| Git Bash | 3 | 0.093, 0.113, 0.127 |
| PowerShell | 2 | ~0, 0.013 |

This is the signature of #1158, which was fixed in v0.7.5 and verified there at ~0.0007 cores/pane. It still reproduces on 0.8.0-preview when panes run Git Bash.

### Expected behavior

Idle panes cost near zero CPU regardless of which shell they run — as they already do with PowerShell.

### Reproduction

1. Windows with `terminal.default_shell` set to Git Bash, server running with idle panes.
2. Measure server CPU over 12 s:
   `$p=Get-Process -Id <server-pid>; $c=$p.TotalProcessorTime.TotalSeconds; Start-Sleep 12; $p.Refresh(); ($p.TotalProcessorTime.TotalSeconds-$c)/12`
3. `herdr workspace create --label t`, then `herdr pane split --pane <root-pane> --direction right --no-focus` ×5. Wait for the prompts to settle.
4. Re-measure: +0.6–0.8 cores.
5. `herdr workspace close <ws>`; CPU returns to baseline within ~0.1 cores.
6. Set `default_shell` to `powershell.exe`, run `herdr server reload-config`, repeat steps 3–5: the same six panes add ~0.08 cores total.

Baseline drifts 3.0–3.5 cores on a live session, so the A-B-A delta is the reliable figure; the before/after baselines agreed within 0.02–0.12 cores in every run.

### Impact

An idle 27-pane session permanently occupies ~3 of 8 cores on this machine, around the clock, with no agent doing any work. It also scales with unrelated system load: spawning 120 idle processes with no new panes added a further ~0.2 cores.

[...]

PR #2494, opened 2026-08-07
fix(windows): reduce idle Git Bash pane CPU

Pimpmuckl:
This addresses the idle Git Bash CPU regression reported in #2459.

## What changed

Idle Git Bash panes were repeating the same process scan for every pane. This shares the reusable part of that scan while keeping the final agent match specific to each pane.

PowerShell panes already avoided this work, which is why the issue only showed up with Git Bash.

Thanks to @hernant for the excellent investigation and proposed approach in #2459.

## Before and after

I reproduced this with isolated Herdr servers and compared one idle pane with seven idle panes over 30-second samples.

| Build | 1 Git Bash pane | 7 Git Bash panes | Added CPU per pane |
| --- | ---: | ---: | ---: |
| Current master | 0.298 cores | 1.889 cores | 0.265 cores |
| This change | 0.286 cores | 0.319 cores | 0.005 cores |

That is about 98% less added CPU per Git Bash pane and 83% less total server CPU with seven panes.

A matching PowerShell control showed no measurable CPU increase per pane. All pane output stayed unchanged during the samples.

## Agent detection

I also launched Codex, Claude Code, and Pi at the same time from separate Git Bash panes. All three were detected in the correct pane, returned to Bash cleanly, and were detected again after relaunch.

## Validation

- `just check`
- 151 Windows tests
- Git Bash and PowerShell idle-pane measurements
- Git Bash launch, exit, and relaunch checks for Codex, Claude Code, and Pi

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -7,7 +7,7 @@ use std::{
     ptr::{copy_nonoverlapping, null_mut},
     sync::{
         atomic::{AtomicU64, Ordering as AtomicOrdering},
-        Arc, LazyLock, Mutex,
+        Arc, LazyLock, Mutex, OnceLock,
     },
     time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
@@ -233,7 +233,7 @@ pub(crate) fn encode_windows_conpty_fallback(key: &crate::input::TerminalKey) ->
 #[derive(Debug)]
 struct CachedProcessSnapshot {
     built_at: Instant,
-    entries: Arc<Vec<WindowsProcessEntry>>,
+    snapshot: Arc<ProcessSnapshot>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


ogulcancelik:
lgtm ty!

hernant:
Tested `7d77e92` on the rig from #2459 (8 cores, ~700 processes — the heavy case). Isolated server, 12 idle Git Bash panes, same methodology as the issue:

| | stock `d78e3d3` | this PR |
|---|---|---|
| total CPU | 1.271 cores | 0.125–0.148 cores |
| IO Other ops/s | ~270,000 | ~26,500 |
| marginal per added pane (A-B-A, +6) | 0.09–0.13 cores | ~0 |

So this fix fully resolves the reported problem here — pane scaling is gone and idle CPU is ~9× lower. For comparison, the dual-cache build measured 0.068 cores / ~146 ops/s on the same setup, so the longer-lived caches buy roughly another 2× CPU on a busy process table — nothing that changes the picture; your call whether it's ever worth the extra state.

One independent observation: most of the remaining ~26K ops/s is `process_runtime_marker`'s 1s negative TTL re-reading environment blocks of the same candidates. Since a process's environment can't change after spawn, caching negative results for the lifetime of `(pid, creation_time)` would close that too, orthogonal to the snapshot-lifetime design choice here.

*Same disclosure as in #2459: testing and analysis done by an AI agent (Claude) on my machine, with me supervising.*

Pimpmuckl:
> One independent observation: most of the remaining ~26K ops/s is process_runtime_marker's 1s negative TTL re-reading environment blocks of the same candidates. Since a process's environment can't change after spawn, caching negative results for the lifetime of (pid, creation_time) would close that too, orthogonal to the snapshot-lifetime design choice here.

Iirc Windows processes can change their env after launch, but you still have a point: Herdr's marker is assigned before the Git Bash starts. It can't appear later. 

Positive marker results are already cached this way.

I'll take a look, thank you for the test!

Edit: Tested this separately. The change is valid, but it did not produce a measurable performance improvement on my rig. Across two A-B-A comparisons with 12 idle Git Bash panes, current master stayed around 28K to 30K IO Other ops/s, and a build containing only this change stayed at the same level.

So the one-second negative TTL does not appear to explain most of the remaining ~26K ops/s on my machine. The much lower result from the earlier dual-cache build cannot be attributed to this part alone because both caches changed at the same time.

This does not mean it saves no environment reads, only that the savings were not material in this test.

I'd say we keep an eye on this and if future profiling warrants another investigation then we reopen the can of worms.

@hernant just tagging you in case github doesn't ping you.

Issue #2356, opened 2026-08-05
Windows auto-update to 0.8.0-preview left release dir empty and symlinks on old version (stuck half-swap)


Qiiks:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

On Windows, updating from 0.7.5-preview to 0.8.0-preview leaves the installation half-swapped:

- `%USERPROFILE%\.herdr\packages\standalone\releases\0.8.0-preview.2026-08-04-...\` is empty (created, never populated).
- `...\releases\.staging.0.8.0-preview.2026-08-04-...15852\` holds the full payload (herdr.exe, conpty/, THIRD-PARTY-NOTICES/).
- Two `.backup.0.8.0-*` and one `.backup.0.7.5-*` directories exist.
- Symlinks `~/.herdr/packages/standalone/current` and `%LOCALAPPDATA%\Programs\Herdr\bin` still point at the 0.7.5 release.
- `herdr --version` reports 0.7.5; `herdr status` reports `restart_needed: no`.

### Expected behavior

The new release directory is populated, the `current`/`bin` symlinks point at 0.8.0, and `herdr --version` reports the new version.

### Reproduction

1. Install 0.7.5-preview.2026-07-29-44b3adb12552 on Windows (preview channel).
2. Run `herdr update` so it downloads 0.8.0-preview.2026-08-04-d78e3d3b5126.
3. Inspect `%USERPROFILE%\.herdr\packages\standalone\releases\` and both symlinks.

Result: staging dir has the payload, the versioned release dir is empty, symlinks unchanged, version still 0.7.5.

### Impact

Windows preview-channel users silently stay on the old version; the binary cannot be updated without manually copying files and repointing symlinks.

### Environment

- Herdr version: 0.7.5-preview.2026-07-29-44b3adb12552 (target: 0.8.0-preview.2026-08-04-d78e3d3b5126)
- Update channel: preview
- Operating system: Windows 10.0.26100.8655
- Terminal: Windows Terminal
- Shell, if relevant: git-bash
- Relevant config, if any: none

Qiiks:
Tried the repro. Two findings:

## 1. Inside a herdr session — `herdr update` refuses (expected guard)

[...]

PR #2530, opened 2026-08-08
fix: make Windows installer swaps atomic

Pimpmuckl:
## Problem

While looking into #2356, I could reproduce the half-swapped update with the exact 0.7.5 and 0.8.0 preview packages.

Windows handles open files differently from POSIX. A file can only be renamed or deleted when every open handle allows delete sharing. That means a short-lived antivirus, indexer, or process handle on the staged package can block the installer swap.

PowerShell's `Move-Item` does not fail atomically for this directory move. It creates the destination directory first, then fails when it reaches the locked file. The installer therefore leaves the full payload in `.staging.*` and an empty versioned release directory.

The rollback only restores the backup when the release directory does not exist. Since the empty directory exists, it skips rollback. Repeated attempts explain the empty `.backup.*` directories from the report while both junctions stay on the old version.

## Fix

Use `System.IO.Directory.Move` for the three release transaction moves:

- current release to backup
- staging to current release
- backup back to current release on failure

These paths are siblings on the same volume, so this uses one native directory rename. Under the same forced file handle it fails before creating the destination, which lets the existing rollback restore the previous release cleanly.

This intentionally does not add retries or new cleanup behavior. If Windows keeps the file locked, the update still reports the failure with a suggestion to be tried, but it no longer leaves the install half-swapped.

## Validation

- reproduced the original state with a real Windows handle that allowed reads but not delete sharing
- added the same handle contention to the existing Windows PowerShell 5.1 installer test
- verified the old `Move-Item` path fails the regression by stranding a backup
- verified the atomic move restores the prior release and the next unlocked attempt succeeds
- `CARGO_BUILD_JOBS=2 just check`

Refs #2356

### DIFF
--- a/scripts/windows_install_conpty_package_test.ps1
+++ b/scripts/windows_install_conpty_package_test.ps1
@@ -133,7 +133,8 @@ try {
         }
     }
 
-    $releaseDir = Get-ChildItem -LiteralPath (Join-Path $herdrHome "packages\standalone\releases") -Directory |
+    $releasesDir = Join-Path $herdrHome "packages\standalone\releases"
+    $releaseDir = Get-ChildItem -LiteralPath $releasesDir -Directory |
         Where-Object { -not $_.Name.StartsWith(".staging.") } |
         Select-Object -First 1
     if ($null -eq $releaseDir) {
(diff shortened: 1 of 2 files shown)


ogulcancelik:
ty!

PR #2535, opened 2026-08-08
feat: make outer pane borders configurable

dhh:
## Summary

- add `ui.pane_outer_borders` with a default of `true` to preserve existing behavior
- allow the outside pane frame to be hidden while retaining shared, draggable split dividers
- apply the setting consistently to visible panes, background resume geometry, and pane layout API output
- document the option in the generated default config and configuration reference

## Configuration

For tmux-style internal splitters without an outside frame:

```toml
[ui]
pane_gaps = false
pane_outer_borders = false
```

### Before: default outer pane frames

Four panes with complete outer frames

### After: internal splitters only

Four panes with only shared internal splitters


ogulcancelik:
ty!

PR #2538, opened 2026-08-09
fix(ui): reclaim scrollbar gutter in alternate screen

ogulcancelik:
## Summary
- hide the pane scrollbar gutter while the terminal uses the alternate screen
- give fullscreen terminal applications the reclaimed column
- restore the stable primary-screen gutter after leaving alternate screen

## Testing
- `just check`
- `just test-one alternate_screen_reclaims_scrollbar_gutter_and_restores_it_on_exit`

### DIFF
--- a/src/ui/panes.rs
+++ b/src/ui/panes.rs
@@ -31,8 +31,11 @@ fn pane_border_title(label: &str, pane_width: u16, _focused: bool) -> Option<Str
     Some(format!(" {} ", truncate_end(label, max_label_width)))
 }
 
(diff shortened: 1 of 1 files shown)


ogulcancelik:
CodeRabbit nitpick: I’m keeping the transition test focused on the shared `terminal_inner_rect` behavior. `compute_pane_infos(..., resize_panes = true)` already proves the active resize and runtime width transition; both `resize_tab_panes` branches delegate directly to the same helper, so separate zoomed/non-zoomed tests would duplicate that contract without covering distinct logic.

PR #2560, opened 2026-08-09
feat: show a zoom indicator in the desktop tab bar

dhh:
## Summary

- show a right-aligned ` ZOOM ` pill in the desktop tab bar while the focused pane is zoomed, styled like the mode bars (accent background, bold contrast text)
- reserve the pill's width in `compute_tab_bar_view` so tabs, scroll buttons, the `+` button, and the overflow ellipsis never render underneath it
- the pill only appears while zoomed, so the tab row is unchanged the rest of the time; the existing per-tab ` Z` label suffix stays for background tabs

The per-tab ` Z` marker is easy to miss when scanning the bar; a fixed-position indicator gives the zoomed state one stable place to look, matching what tmux users get from `#{?window_zoomed_flag,ZOOM ,}` in status-right.

## Testing

- new unit tests: pill renders at the right edge with reserved layout space when the active tab is zoomed, and is absent otherwise
- existing tab bar, mode bar, and full-frame characterization tests pass unchanged


ogulcancelik:
superseded by https://github.com/herdrdev/herdr/pull/2586 with the new `ui.tab_bar_right`

PR #2600, opened 2026-08-10
fix: throwaway repros silently used the live session

HackAttack:
The herdr-throwaway-repro skill did not have its intended effect. A disposable session never started, and once it failed there was nothing left to stop the reproduction from running in the user's live session.

The skill's only documented way to create a disposable session was `herdr --session <name>` typed into an outer pane. That launches the TUI, and a TUI launched from a Herdr-managed pane exits 1 with "nested herdr is disabled by default" unless the user opted into `experimental.allow_nested`.

The rejection was printed inside the scratch pane, which was closed later, while the driving shell only saw 25 seconds of "No such file or directory" from polling for a session socket that was never going to appear. Then it fell back to `herdr pane send-text/send-keys/read` with no explicit session and no env prefix, which hit the default session.

Fix by starting the disposable session with `herdr --session <name> server`, which is dispatched before the nested gate and needs no config change or outer pane. Address it with the `--session` flag, which marks the session explicit so the inherited HERDR_SOCKET_PATH is ignored and a missing session fails with server_not_running. The HERDR_SESSION variable the skill used for targeting does not do this; it loses to that inherited socket path and answers from the user's session instead.

### DIFF
--- a/.agents/skills/herdr-throwaway-repro/SKILL.md
+++ b/.agents/skills/herdr-throwaway-repro/SKILL.md
@@ -7,15 +7,16 @@ description: Create and control a disposable named Herdr session from inside an 
 
 Use a disposable named Herdr session when a reproduction needs a real Herdr server, panes, PTYs, agents, or socket API without risking the user's main session.
 
-The temporary TUI only keeps the disposable session attached and supplies terminal geometry. Drive the reproduction from the parent session through Herdr's CLI/API. Do not manually operate the nested TUI unless the bug specifically requires client input.
+Run that session as a headless server and drive it from the current session through Herdr's CLI/API. A headless server spawns real PTYs with no client attached, so most reproductions need no nested TUI and no extra pane in the user's session.
 
 ## Non-negotiable safety
 
 - Never run the reproduction in the default session.
+- If the disposable session cannot be started or addressed, stop and report. Never continue the reproduction in the current session, and never edit the user's `config.toml` to work around it.
 - Never stop, restart, delete, or kill the main Herdr server.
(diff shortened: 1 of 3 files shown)


ogulcancelik (dismissed):
the throwaway-repro correction is useful, but please keep this PR to .  tracks the latest stable release and cannot change in normal feature/fix work, and  is for user-facing runtime changes rather than project skill maintenance. after dropping those two edits, please rebase onto current master and rerun the checks.

ogulcancelik (changes requested):
the throwaway-repro correction is useful, but please keep this PR to `.agents/skills/herdr-throwaway-repro/SKILL.md`. `skills/herdr/SKILL.md` tracks the latest stable release and cannot change in normal feature/fix work, and `docs/next/CHANGELOG.md` is for user-facing runtime changes rather than project skill maintenance. after dropping those two edits, please rebase onto current master and rerun the checks.

HackAttack:
Got it, done.

JJLiebig:
Not needed anymore on current master head as the Skill got partially reworked.

HackAttack:
This or something like it is still needed indeed, as I just was working on master and not only did my current session get used again, but this time it entirely locked up and I had to forcibly stop it, killing my processes. Please reopen.

JJLiebig:
@HackAttack hmm apologies, curious how that can still happen, my agents haven't really used the main session in ever

Will reopen and take a look at what you suggest

Note that upstream skill has the wording on disposable sessions and using a custom config that sets the nesting to be allowed. In my personal testing mostly with Sol and Astra, this has lead to the expected outcomes so far.

HackAttack:
Thank you, I will try to reproduce deterministically soon.

PR #2614, opened 2026-08-10
feat: remember agent cli options when resuming sessions

dhh:
Native agent session restore rebuilds the resume command from the session reference alone, so a pane started as `claude --permission-mode bypassPermissions` comes back as plain `claude --resume <id>` with default permissions. The same applies to `codex -s danger-full-access -a never` and any other option a pane was started with.

Flagging the process point first: this touches persistence and resume behavior rather than being a focused bug fix, so per CONTRIBUTING.md it is the kind of change that wants maintainer alignment. Happy to move it to a Discussion if you would rather shape the approach before looking at code.

## What changed

**Capture** — the foreground process probe already reads each process's argv, so when it identifies an agent it also extracts the command line after the agent token (`detect::agent_launch_args_in_job`, which handles wrapped invocations like `node .../cli.js --flag`). A new `AgentLaunchArgsDetected` event records it on the terminal, deduped so it fires once per change rather than once per probe. No new process inspection or filesystem work is added to the detection loop; it reuses the argv the probe already has.

**Persist** — a new optional `agent_launch: {agent, args}` field on the pane snapshot, stored per agent so a pane that later ran a different agent never resumes with the previous agent's options. Old snapshots deserialize with the field absent.

**Replay** — `agent_resume::plan_with_launch_args()` appends the surviving options after the session reference. `src/agent_launch_args.rs` decides what survives, with two rules:

- Options that select a conversation or a one-shot run are dropped, per agent (`--resume`, `--continue`, `--session`, `--print`, `--prompt`). Herdr supplies its own session reference.
- A bare word is kept only when it directly follows a kept option, where it is that option's value. Bare words elsewhere are prompts, subcommands, or positional paths.

Everything else is replayed, including options Herdr has never heard of, so a new flag in a future agent release survives a resume without anyone touching Herdr.

The second rule is what makes the first cheap. The value of a dropped option disappears for free, because it is a bare word following no kept option: `--resume <id>` loses the id without Herdr knowing that `--resume` takes a value. The same falls out for agents that select a session through a subcommand and a positional id, so `codex resume <id> --full-auto` needs no special case. That means the only per-agent data is a list of option names, with no arity table to drift as the agent CLIs change.

It also stops a resume from compounding: after Herdr resumes a pane once, the live process argv contains the `--resume <id>` Herdr itself appended, and the next restore drops it before adding the current one.

## Deliberate limits

- An option given several separate values keeps the first: `--add-dir /a /b` resumes as `--add-dir /a`.
- A prompt written directly after a flag cannot be told apart from that flag's value, so it is replayed. `claude "fix the bug"` and a prompt following a value are both dropped; `claude --dangerously-skip-permissions "fix the bug"` is not.

Both are documented in the module header and on the session-state page. Distinguishing those cases needs a per-agent table of every option and its arity; an earlier revision had one and it was not worth its weight.

## Verification

`cargo nextest run` passes (3376 tests). `cargo fmt --check` is clean. Unit tests cover the argument rules, the argv extraction including wrapped invocations, the event recording, the snapshot capture, and the restore plan including the per-agent guard.

Also verified live in a throwaway session on a debug build, since unit tests do not cover the probe to snapshot to resume chain:

- a pane agent started as `claude --permission-mode bypassPermissions` recorded `agent_launch: {agent: "claude", args: ["--permission-mode", "bypassPermissions"]}` in `session.json`
- a snapshot seeded with session ref `verify-session-id` and args `["--model", "sonnet", "--resume", "stale-session", "some prompt"]` resumed as `hermes --resume verify-session-id --model sonnet` — option replayed, stale session flag and its id dropped, trailing prompt left behind

Two notes on `just ci` from my machine: `clippy -D warnings` fails on two pre-existing lints in `src/server/handoff.rs:381` and `src/terminal_theme.rs:136`, both files untouched by this change and both flagged only because my toolchain is newer than the pinned 1.96.1. `integration-assets-test` and `plugin-marketplace-test` need bun, which I do not have installed; neither covers this change.

### DIFF
--- a/docs/next/website/src/content/docs/session-state.mdx
+++ b/docs/next/website/src/content/docs/session-state.mdx
@@ -81,6 +81,22 @@ Native session restore requires these Herdr integration versions or newer:
 | Hermes Agent | `2` | `hermes --resume <id>` |
 | MastraCode | `1` | `mastracode --thread <id>` |
 
+Herdr also replays the options the agent was started with, so a pane started as
+`claude --permission-mode bypassPermissions` resumes as
+`claude --resume <id> --permission-mode bypassPermissions` instead of dropping back to
+default permissions.
+
+Options that select or continue a conversation (`--resume`, `--continue`, `--session`) and
+options that switch the agent to a one-shot run (`--print`, `--prompt`) are never replayed;
+every other option is, including ones Herdr has never heard of. A bare word is replayed only
+when it directly follows a replayed option, where it is that option's value, so prompts,
+subcommands, and positional session ids are left behind.
+
+Two limits follow from that rule: an option given several separate values keeps only the
+first (`--add-dir /a /b` resumes as `--add-dir /a`), and a prompt written directly after a
+flag cannot be told apart from that flag's value, so it is replayed. Use `--option=value`
+form if you want to be sure a following prompt is not mistaken for a value.
+
 Run `herdr integration status` to check installed integration versions. Reinstall outdated integrations with `herdr integration install <agent>`.
 
 Unsupported, missing, invalid, duplicated, or stale session references restore as normal shells in the saved pane directory.
--- a/src/agent_launch_args.rs
+++ b/src/agent_launch_args.rs
@@ -1,0 +1,287 @@
+//! Remembering the options an agent CLI was started with.
+//!
+//! Herdr resumes a native agent session by running the agent again with its own
+//! session reference. The rest of the original command line matters too: a pane
+//! started with `claude --permission-mode bypassPermissions` must come back with
+//! that permission mode, not with plain `claude --resume <id>`.
+//!
+//! The captured argument list comes from the live agent process, so it can also
+//! contain the session reference Herdr appended on a previous resume, a
+//! subcommand, or the prompt the user typed at launch. Replaying those would
+//! either point the agent at a stale conversation or restart work the user never
+//! asked for, so two rules decide what survives a resume:
+//!
+//! - Options that select a conversation or a one-shot run are dropped, per
+//!   agent. Herdr supplies its own session reference.
+//! - A bare word is kept only when it directly follows a kept option, where it
+//!   is that option's value. Bare words elsewhere are prompts, subcommands, or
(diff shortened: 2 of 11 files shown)


dhh:
Good catch, fixed in 29eba781.

I went slightly wider than the suggested diff on two points:

1. **Same bug in a second place.** The `was_imported` branch further down (`restore.rs`, the `match runtime_result` arm) stored `saved_agent_launch` unfiltered too. Fixing only the native-restore branch would have left that path re-persisting a mismatched agent's options.

2. **`is_some_and` would drop too much.** The proposed `saved_agent_session.is_some_and(...)` discards the saved options whenever no session was recorded. That is harmless at the native-restore site, which only runs when a restore plan exists and therefore always has a session, but at the imported site a session is not guaranteed, so it would throw away options that nothing contradicts.

Both sites and `restore_plan_for_snapshot` now share one helper, `agent_launch_for_session`, which keeps the options unless a recorded session names a different agent. Covered by `saved_agent_options_survive_only_a_matching_or_absent_session`.

dhh:
Fixed in 2fa655cf.

Confirmed: with unreadable argv the early return left the previous invocation's options in `TerminalState`, and since resume matches on the agent label alone, a later run of the same agent could resume with its predecessor's model, permission, sandbox, or approval flags.

The probe already knew which job it was looking at, so `last_reported` is now keyed on the tracked process group as well as the agent. Unreadable options are handled by which job is in front:

- same job as the last report: a transient read failure, so the last options stand;
- a different job: the stored options are its predecessor's, so empty options are published and nothing stale is replayed;
- nothing reported yet: left alone, since anything stored came from a restored snapshot and is not the probe's to overwrite.

That last case matters — clearing unconditionally would have wiped the options this PR restores from the snapshot before the first successful probe.

Tests: `unreadable_options_never_carry_across_agent_invocations` and `unreadable_options_leave_restored_ones_alone`.

ogulcancelik:
i'm working on a new agent resume manifest system which is going to supersede this, so closing for now.

PR #2627, opened 2026-08-10
feat: keep the outer terminal window title in sync with the session

dhh:
## Problem

Herdr emulates the terminals in its panes, so an `OSC 0`/`OSC 2` title written inside a pane stops at Herdr and never reaches the terminal Herdr itself runs in. That outer title is what window managers and terminal tab bars read, so it keeps showing whatever the shell or `ssh` happened to leave behind.

It shows up most clearly over SSH. `tmux` has `set-titles on` / `set-titles-string`, so an SSH'd tmux session labels the window with the remote host and window name and a tiled window manager's tab or group bar follows along. Herdr has no equivalent — the only thing that ever writes an outer title is the manual `client.window_title.set` API.

## Change

Adds `ui.window_title`, a format string rendered against current session state and pushed to the foreground client whenever it changes:

```toml
[ui]
window_title = "{hostname}: {workspace}"   # the default
```

Tokens are `{hostname}`, `{workspace}`, `{tab}`, `{pane}` (the focused pane's manual name), and `{terminal_title}` (the focused pane's own terminal title with spinner frames stripped). `{{`/`}}` are literal braces, an unknown token is a config diagnostic that leaves the outer title alone, and `window_title = ""` disables the feature.

Notes on the design:

- **It renders on the server.** `{hostname}` names the machine the panes actually run on, so it stays correct for `herdr --remote` thin clients and for running `herdr` over SSH — same semantics as tmux's `#h`. This also reuses the existing `ServerMessage::WindowTitle` path that `client.window_title.set` already goes through, rather than adding a second one.
- **It costs nothing when idle.** The sync is gated on a pending render, since anything that can move the title also asks for a render, and the hostname is resolved once when config is applied rather than per render. It is O(1) per loop, not pane-scaled.
- **A title is only remembered as delivered once a foreground client takes it**, so the first client to attach is written to rather than skipped. (This was a real bug in an earlier revision of this branch: the server renders once before any client attaches, so recording that as sent meant the actual terminal never got a title.)
- `client.window_title.set` still wins over the configured title. Clearing it now hands the title back to `ui.window_title` instead of only `"herdr"`.

## Verification

`just check` equivalent locally: `cargo fmt --check`, clippy, 3392 nextest tests, and the maintenance script suites all pass. Six new tests cover template parsing, rendering, the once-per-change push, the API override and hand-back, first-attach delivery, and client promotion.

Also driven end to end against a real PTY on a throwaway session, reading the OSC bytes off the wire while switching and renaming the active workspace:

```
OSC0 titles: [b'host: herdr', b'host: omarchy', b'host: prod deploy']
```

One emission per change, and nothing while idle.

## Docs

`docs/next` configuration guide (English, ja, zh-cn), the config reference data, the generated default config comments, the changelog, and the `client.window_title.clear` wording in the CLI and socket API docs.


dhh:
@coderabbitai review

dhh:
Fixed in ce865898.

`send_to_client` returns `true` whenever the client entry exists, and only the inner `if let Some(writer)` guards the actual queueing, so a detached client with `writer: None` reported a successful send and got cached in `sent_window_title`. That defeated the very thing the second commit here added: on reattach the title matched the cache and was skipped, leaving the new terminal with a stale title.

`send_window_title` now requires a writer before sending and clears the cache otherwise, matching the existing no-foreground-client branch. Regression test: `a_foreground_client_without_a_writer_does_not_cache_the_window_title`, which fails on the previous code with `assertion failed: server.sent_window_title.is_none()`.

dhh:
Fixed in 90fb47a5.

You were right that the replacement server started with `api_window_title: None`, so an override set through `client.window_title.set` was silently replaced by the configured `ui.window_title` on the first render after an update.

`HandoffManifest` now carries `api_window_title`, the outgoing server populates it, and the import applies it before `report_ready`, so the override is in place before any client reattaches. The field is `#[serde(default)]` and the manifest is JSON, so a manifest written by a server that predates the field still loads with `None` and needs no `HANDOFF_VERSION` bump.

Tests: `a_handoff_carries_an_api_set_window_title` and `a_manifest_written_before_the_title_field_still_loads`.

ogulcancelik:
Addressed CodeRabbit’s terminal-title sync performance note in 9e0e5272. API requests now snapshot only pending title-source pane IDs, and title synchronization resolves those panes directly instead of scanning and locking every pane runtime.

ogulcancelik:
this was on backlog for a long while. ty!

PR #2647, opened 2026-08-11
fix: make just build cross-platform

Pimpmuckl:
## What changed

`just build` currently uses the Unix shell on Windows. It prints the Cargo command, then fails when `sh` is not installed.

This keeps the existing Unix recipe as-is and adds a Windows variant using the same PowerShell recipe pattern already used by `lint` and `check`.

No extra script needed.

## Testing

- `just build` on Windows
- incremental `just build`
- `just --show build`
- `just lint`

### DIFF
--- a/justfile
+++ b/justfile
@@ -60,6 +60,12 @@ install-hooks:
     @echo "installed git hooks from .githooks"
 
 # Build release binary
+[unix]
+build:
+    cargo build --release --locked
+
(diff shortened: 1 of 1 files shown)


Pimpmuckl:
The mc CI failure looks like a flake. This PR does not touch that code, and the same test passed on the base commit shortly before this run. The test relies on asynchronous task cancellation and appears to have hit a timing race under CI load.

@kangal-bot could you trigger a rerun perchance?

Issue #2643, opened 2026-08-10
Windows: Git sidebar polling repeatedly triggers Defender scans for WSL repositories


Pimpmuckl:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

An idle Herdr workspace whose cwd is a Git repository under `\\wsl.localhost` keeps triggering Microsoft Defender scans even when the repository does not change.

In a 30-second trace attributed to the exact isolated Herdr server PID:

| Condition | Defender scans caused by Herdr | `.git/config` scans |
| --- | ---: | ---: |
| Local NTFS repo, default sidebar | 0 | 0 |
| WSL UNC repo, default sidebar | 162 / 719.750 ms | 108 / 660.750 ms |
| Same WSL repo without Branch/GitStatus rows | 0 | 0 |

The 108 config scans came in 18 bursts of exactly six, with a median interval of 1.607 seconds.

I also captured this in my normal session over 90 seconds: Herdr caused 395 Defender scans taking 22.59 seconds. The WSL repository config accounted for 252 scans and 22.27 seconds, including one scan that took 19.76 seconds.

### Expected behavior

Showing Branch and GitStatus should not make an unchanged WSL repository trigger continuous Defender scans. Once the repository is warm and unchanged, Herdr should avoid repeatedly reopening the same Git configuration every couple of seconds.

### Reproduction

1. Build upstream/master at `42789c8e1d86a3d5ea529b1b047172bfd6979a21` on Windows.
2. Create a Herdr workspace whose cwd is a Git repository under `\\wsl.localhost\<distribution>\home\<user>\<repo>`.
3. Keep the default Branch and GitStatus rows enabled and leave the workspace idle.
4. From an elevated PowerShell, record 30 seconds:
   `New-MpPerformanceRecording -RecordTo "$env:TEMP\herdr-defender.etl" -Seconds 30`
5. Inspect the result:
   `Get-MpPerformanceReport -Path "$env:TEMP\herdr-defender.etl" -TopProcesses 20 -TopFiles 20 -TopScans 20`
6. Repeat with Branch and GitStatus removed from the spaces sidebar.

The default-sidebar run repeatedly scans the WSL repository's `.git/config`. The no-Git-sidebar control produces no scans attributed to Herdr.

### Impact

[...]

PR #2688, opened 2026-08-11
fix: avoid repeated git config reads 

Pimpmuckl:
## What changed

Git status refreshes now keep the repository layout and branch config in the existing cache.

The 1.5-second refresh still reads HEAD, the current branch ref, and the upstream ref, so branch and ahead/behind stay live. Config is re-read when the cwd, branch, worktree layout, ref storage, or a tracked config input changes. There is still a full five-minute reconciliation for filesystem timestamp collisions.

The existing two-pass include handling also shares raw file contents during one rebuild, so the same physical config file is only opened once per rebuild.

## Why

On Windows, opening Git config over a WSL UNC path can trigger Defender scanning. I could recreate this on a private WSL repo and traced six `.git/config` reads per refresh.

A metadata-only control caused zero Defender scans, so this checks file metadata on the fast path instead of reopening config contents. This keeps the Git sidebar behavior while removing the expensive part.

## Before and after

Matched 20-second warmup and 30-second exact-server-PID Defender traces used the same checkout with the default Branch and GitStatus fields. The baseline was master after #2651.

| | Baseline | This PR |
|---|---:|---:|
| `.git/config` scans | 114 | 0 |
| Defender scan time | 679 ms | 0 ms |
| Process read bytes | 3,100,789 | 2,172 |

Both runs showed a private sample repo on the right `main` branch correctly.

## Testing

- `just check`
- focused workspace Git tests, 51 passed
- Git refresh tests, 12 passed
- `cargo clippy --bin herdr -- -D warnings`

refs #2643

### DIFF
--- a/src/app/git_refresh.rs
+++ b/src/app/git_refresh.rs
@@ -141,6 +141,7 @@ fn deduplicate_git_refresh_items(
     let mut jobs = Vec::<WorkspaceGitRefreshJob>::new();
 
     for item in items {
+        let reconcile = item.cache_key_hint.is_none();
         let cache_key = item.cache_key_hint.unwrap_or_else(|| {
             crate::workspace::git_status_cache_key(&item.resolved_identity_cwd)
                 .unwrap_or_else(|| item.resolved_identity_cwd.clone())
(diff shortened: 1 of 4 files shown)


ogulcancelik:
ty!

PR #2722, opened 2026-08-12
fix: use the running binary in agent hooks

Pimpmuckl:
## What

Agent hooks that call the Herdr CLI now use `HERDR_BIN_PATH`, with `herdr` on `PATH` as a fallback.

This fixes the CLI callback for:

- Claude Code
- Codex
- Kimi
- GitHub Copilot CLI
- Droid
- Qoder CLI on Windows and Unix
- Antigravity CLI

The other supported integrations already use `HERDR_BIN_PATH` or report through the socket directly.

## Why

I hit this while dogfooding a current master build on Windows. The running server used the current protocol, but the installed `herdr` on `PATH` was older. Codex started normally, but its session hook called the older client and the report failed silently.

That meant Herdr could detect the agent, but never saved its thread ID. After a reboot, panes with older saved IDs resumed correctly while newer panes came back as shells.

Herdr already gives every pane the exact running binary path. The hooks now use it consistently, so their client always matches the server.

## Testing

- `just check`
- PowerShell and shell syntax checks
- direct Codex hook callback with an injected `HERDR_BIN_PATH`

### DIFF
--- a/src/integration/assets/antigravity_cli/herdr-agent-state.ps1
+++ b/src/integration/assets/antigravity_cli/herdr-agent-state.ps1
@@ -2,7 +2,7 @@
 # managed by herdr; reinstalling or updating the integration overwrites this file.
 # add custom hooks beside this file instead of editing it.
 # HERDR_INTEGRATION_ID=antigravity_cli
-# HERDR_INTEGRATION_VERSION=1
+# HERDR_INTEGRATION_VERSION=2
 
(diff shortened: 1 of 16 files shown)


ogulcancelik:
hmm, i really need to work on agent registry v1. we have so much problems on what an agent means inside herdr, how to update or add new ones and now integrations share same versions across operating systems which is bad :/

Issue #796, opened 2026-06-25
Mouse click to switch tabs/workspaces does not update UI on macOS + Ghostty (cell_width_px=0)


mattstratton:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

Clicking on tabs, workspaces, or other herdr UI elements on macOS + Ghostty registers the click in the server log (tab.focus / workspace.focus with outcome="ok") but the UI does not update to show the switched tab or workspace. 

Mouse navigation works intermittently at best (~50% success rate).

Rapid repeated clicking sometimes causes tabs to cycle uncontrollably (1→2→3→1→2→3 at ~17ms intervals) without user input at that speed.

Server logs show cell_width_px=0 cell_height_px=0 on every client connect and resize event:
`client connected client_id=2 cols=149 rows=62 cell_width_px=0 cell_height_px=0 render_encoding=SemanticFrame`

This persists across reconnects and window resizes.

### Expected behavior

Clicking a tab or workspace in the herdr UI should reliably switch to it and redraw the content.

### Reproduction

  1. Run herdr in Ghostty on macOS
  2. Create 2+ tabs or workspaces
  3. Click a different tab or workspace in the herdr UI bar
  4. Server log shows tab.focus outcome="ok" but displayed content does not change
  5. Repeat — succeeds inconsistently

### Impact

Mouse navigation in herdr is essentially unusable on macOS + Ghostty. Keyboard shortcuts (ctrl+b + n/p) work as a workaround but all mouse-based
  tab and workspace switching is broken.

### Environment

  - Herdr version: 0.7.1
  - Update channel: stable
  - Operating system: macOS Darwin 25.5.0
  - Terminal: Ghostty 1.3.1 (stable)
  - Shell: zsh
  - Relevant config: stock herdr config (no `mouse_capture` override); stock ghostty config (no mouse-related overrides, `mouse-reporting = true` default)

ogulcancelik:
can’t reproduce this locally, and i can’t reproduce it in other setups either.

please try the same flow in another terminal, with another mouse or trackpad, and with any ghostty mouse/input config disabled. right now this looks more like a local terminal/input-device issue than an actionable herdr bug.

if you can share enough repro data ping me i'll reopen but at it's current state this doesn't look like herdr.

mattstratton:
Thanks for the response. Here's the specific repro data from the server log:

[...]

PR #2736, opened 2026-08-12
fix: keep tab and sidebar clicks when the terminal reports a stray drag

HackAttack:
This works around a Ghostty bug that causes herdr tabs to be unclickable in fullscreen mode.

Ghostty on macOS in native fullscreen emits one motion report after a fresh press in the strip where the menu bar reveals, with the row shifted to near the bottom of the grid. The cause is upstream: macOS delivers one mouse event whose y is nearly a window height away from the real pointer, and Ghostty reports it as it arrives (iTerm2 and Terminal.app report the same clicks correctly).

Herdr treated the stray report as a real drag and lost the click two ways: it opened a tab reorder with no drop target, making the release a no-op, and it handed the release to the pane under the bogus coordinates.

Now, a press on the tab bar or sidebar owns the gesture until release, and a tab drag only starts once the pointer is over a real drop target.

refs #796


HackAttack:
Thanks for the assist! I had not considered multiple concurrent clients.

PR #2777, opened 2026-08-13
feat: persistent command layer (vim-like navigation mode)

adrien-barret:
## What

Adds a persistent, vim-like **command layer** (`Mode::FocusNav`) — a normal-mode
counterpart to the terminal's insert mode. Enter it via a configurable `focus_nav`
keybind and **the whole prefix keymap goes live on bare keys**, staying open across
actions instead of exiting after one:

- Movement and actions use your existing prefix bindings without the prefix:
  `h/j/k/l` focus panes, `c` new tab, `v` / `-` split, `p` / `n` switch tabs,
  `w` workspace picker, `x` close, `z` zoom, and so on.
- The mode **stays open** after each action — chain as many as you like.
- **Escape** (or re-pressing the trigger) exits back to the terminal.

A `CMD` status bar is shown while the layer is active.

### Text dialogs suspend and return

Actions that open a text input (rename workspace/tab/pane, new-tab / new-workspace
name prompts, confirm-close) **suspend** the command layer so you can type. On
**commit or cancel** you return to the command layer — not to the terminal — so an
accidental rename never drops you out of navigation.

## Why

Today, movement in the live layout is one-shot: `prefix+h/j/k/l` moves focus once
and returns to the terminal, and `goto` / `workspace_picker` are pickers that close
on selection. There is no sticky, modal layer to drive herdr with bare keys. This
adds a familiar normal/insert split over panes, tabs, and workspaces, reusing the
existing keymap so there is nothing new to learn.

## Design

- New `Mode::FocusNav`; `handle_focus_nav_key` dispatches the same prefix keymap
  (`non_indexed_action_for_key` / `command_for_key` / `indexed_navigation_action`
  with `BindingDispatch::Prefix`) but never auto-exits.
- New `AppState.command_layer` flag, set on entry and kept across dialogs. The
  shared return helpers (`leave_navigate_mode`, `leave_command_mode`, `leave_modal`,
  `confirm_close_cancel`) route back to `Mode::FocusNav` while the flag is set — this
  is what makes dialogs suspend-and-return.
- New bindable action `EnterFocusNav`, wired as `focus_nav`, **unset by default** —
  no change to existing keymaps. Existing `Navigate` / `Navigator` behavior is
  untouched.

## Config

```toml
[keys]
focus_nav = "prefix+space"   # unset by default
```

## Testing

- `nix build .#herdr` passes.
- `herdr config check` accepts `focus_nav` and still flags unknown keys.
- Manual: enter the layer, chain pane/tab/workspace actions, open a rename dialog
  and confirm both commit and Escape return to the command layer; Escape from the
  layer returns to the terminal.

### DIFF
--- a/src/app/input/mod.rs
+++ b/src/app/input/mod.rs
@@ -107,6 +107,7 @@ impl App {
                 Mode::OpenExistingWorktree => self.handle_worktree_open_key(key_event),
                 Mode::ConfirmRemoveWorktree => self.handle_worktree_remove_key(key_event),
                 Mode::Resize => self.handle_resize_key_via_api(key),
+                Mode::FocusNav => self.handle_focus_nav_key(key),
                 Mode::ConfirmClose => self.handle_confirm_close_key_via_api(key_event),
                 Mode::ContextMenu => {
                     self.handle_context_menu_key_via_api(key_event);
--- a/src/app/input/modal.rs
+++ b/src/app/input/modal.rs
@@ -1158,6 +1158,65 @@ impl App {
         }
     }
 
+    /// Persistent focus-navigation mode (like resize mode, but moves focus).
+    /// Stays open so you can chain moves; esc/enter or re-pressing the trigger exits.
+    ///   h/j/k/l or arrows -> focus pane in that direction
+    ///   Tab / Shift+Tab   -> next / previous tab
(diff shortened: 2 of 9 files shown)


adrien-barret:
@coderabbitai review

Issue #2751, opened 2026-08-13
[Windows] herdr update triggers Cortex XDR downloader rule and terminates Windows Terminal


kurtontheway:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug twice on the version and environment reported below using the exact steps provided.

### Current behavior

Running `herdr update` on Windows triggers Palo Alto Cortex XDR rule `bioc.malicious_downloader_ps_webr`. Cortex XDR blocks the causality chain and explicitly terminates `WindowsTerminal.exe`, closing the complete Windows Terminal window rather than only failing the update command.

The process chain recorded by Cortex XDR is:

```text
WindowsTerminal.exe
  -> pwsh.exe
    -> herdr.exe update
      -> powershell.exe -NoProfile -ExecutionPolicy Bypass
         -Command "irm https://herdr.dev/install.ps1 | iex"
```

The XDR event records:

```text
rule: bioc.malicious_downloader_ps_webr
action: block
external_description: Windows command was used to download a malicious file
prevention action: Terminate Causality
CGO: WindowsTerminal.exe
```

The prevention reports identify `WindowsTerminal.exe` as the source process with `Terminate="true"` and `TerminateMode="Process"`.

This occurred twice on 2026-08-12 (Asia/Shanghai):

| Local time | Windows Terminal PID |
| --- | ---: |
| 16:04:12 | 19436 |
| 16:05:34 | 38884 |

There were no Windows Error Reporting crash dumps or Application Error events for Windows Terminal. The window was terminated by Cortex XDR rather than crashing.

The second attempt downloaded and extracted the target release before termination:

```text
%USERPROFILE%\.herdr\packages\standalone\releases\
  .staging.0.8.0-preview.2026-08-04-d78e3d3b5126-x86_64-pc-windows-msvc.41528\
```

The staging directory contains `herdr.exe`, `conpty\`, and third-party notices. However, `%LOCALAPPDATA%\Programs\Herdr\bin` still points to the old 0.7.4-preview release, so the update remains incomplete.

The staged executable remains present. This specific detection was a behavioral downloader rule against the PowerShell execution chain, not a binary quarantine event.

### Expected behavior

`herdr update` should complete without invoking a command pattern that endpoint protection classifies as a malicious PowerShell downloader, and it should never cause the user's terminal host to be terminated.

[...]

PR #2816, opened 2026-08-14
fix: use verified local packages for windows updates

Pimpmuckl:
## What

Windows self-updates still ran the web installer through:

```powershell
irm https://herdr.dev/install.ps1 | iex
```

That process chain triggers Cortex XDR's PowerShell downloader rule. In #2751 it terminated the complete Windows Terminal causality chain and left the update incomplete.

`herdr update` now downloads the exact Windows package selected by the already-fetched release manifest, verifies its published SHA-256, and runs the bundled installer against that local package.

This brings Windows in line with the existing Linux and macOS update flow for release acquisition. Herdr now owns manifest selection, artifact download, and checksum verification on every platform. PowerShell receives only local paths and release metadata.

The public `install.ps1` flow stays unchanged for first-time installs.

## Why

Linux and macOS already let Herdr select the release, download the exact artifact, verify it, and then perform the platform-specific activation.

Windows took a separate detour. After Herdr selected the release, it handed acquisition back to a remote PowerShell script, which fetched the manifest and package again. That duplicated work and looked exactly like a PowerShell downloader to endpoint protection.

Removing that Windows-only detour simplifies updater ownership and removes the observed trigger. This is not an EDR-specific retry or exception.

The Windows-specific activation path stays where it already works. The PowerShell installer still owns ConPTY validation, the install lock, atomic activation, rollback, junction updates, retention, and cleanup. Only release acquisition moves into Herdr.

## Testing

- `just check`
- Existing Windows installer integration with the checked-in preview package
- Local package checksum rejection and successful activation
- Existing repair, rollback, retention, and junction coverage
- Disposable process trace confirmed only `curl` performs network downloads and PowerShell receives local paths and metadata only

Refs #2751

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ include = [
     "assets/sounds/*",
     "docs/next/api/herdr-api.schema.json",
     "skills/herdr/SKILL.md",
+    "website/install.ps1",
     "README.md",
     "LICENSE",
     "Cargo.toml",
--- a/scripts/windows_install_conpty_package_test.ps1
+++ b/scripts/windows_install_conpty_package_test.ps1
@@ -113,10 +113,56 @@ try {
         }
     }
 
-    & "$PSScriptRoot\..\website\install.ps1" `
-        -ManifestUrl $manifestUrl `
-        -InstallDir $installDir `
-        -ExpectedBuildId "installer-test"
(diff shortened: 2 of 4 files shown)


ogulcancelik:
great ty!

PR #2825, opened 2026-08-14
feat(keybindings): allow configured double-prefix actions

drakejwong:
## Summary

- allow an explicit user keybinding whose prefix-mode key equals `keys.prefix`
- dispatch configured prefix actions before falling back to literal double-prefix passthrough
- preserve Escape precedence and the existing passthrough behavior when no override is configured

This enables configurations such as:

```toml
[keys]
prefix = "ctrl+s"
zoom = "prefix+ctrl+s"
```

## Tests

- `cargo fmt --check`
- `cargo test --bin herdr double_prefix`
- `cargo test --bin herdr escape_precedes_configured_prefix_action`
- `cargo test --bin herdr config::keybinds::tests`
- `cargo test --bin herdr app::input::navigate::tests`

### DIFF
--- a/src/app/input/navigate.rs
+++ b/src/app/input/navigate.rs
@@ -65,16 +65,6 @@ impl App {
             return;
         }
 
-        if self.state.is_prefix_key(&raw_key) {
-            if self.state.copy_mode_pane_is_focused() {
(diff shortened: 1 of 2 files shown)


pre:
Related discussion
* https://github.com/herdrdev/herdr/discussions/862

Issue #2916, opened 2026-08-17
Windows ARM64 installer fails to activate x64 release


travis-mcdevitt:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

After the ARM64 support added in #897, the installer detects Windows ARM64 and downloads the x64 preview release. It then fails to activate the release:

> Exception calling "Move" with "2" argument(s): Access to the path '.staging...x86_64-pc-windows-msvc...' is denied.

This reproduced on two installation attempts.  

### Expected behavior

The downloaded x64 release is activated and Herdr installs successfully.

### Reproduction

On Windows ARM64, run:

powershell -ExecutionPolicy Bypass -c "irm https://herdr.dev/install.ps1 | iex"

### Impact

The documented installation command does not install Herdr on my Windows ARM64 machine.

### Environment

- Herdr version: 0.8.0-preview.2026-08-17-1147e60bc0a4
- Update channel (stable or preview): preview
- Operating system: Windows 11 (ARM64)
- Terminal: Windows Terminal 1.24.11911.0
- Shell, if relevant: Windows PowerShell 5.1.28000.2704
- Relevant config, if any: none

travis-mcdevitt:
rename probe: failed
System.Management.Automation.MethodInvocationException
Exception calling "Move" with "2" argument(s): "Access to the path 'C:\Users\[redacted]\.herdr\packages\standalone\releases\.staging.0.8.0-preview.2026-08-17-1147e60bc0a4-x86_64-pc-windows-msvc.4060' is denied."

PR #2921, opened 2026-08-18
fix: retry Windows installer activation after transient locks

ogulcancelik:
Fixes the Windows ARM64 installer activation failure reported in #2916.

The Windows installer now retries the atomic staging-directory activation while Windows releases transient executable handles. A native `windows-11-arm` workflow verifies that the published x86_64 preview package installs and launches under emulation.

The diagnostic workflow reproduced the published installer failure, and the fixed installer passed three consecutive native ARM64 runs before the regression test was finalized.

Refs #2916

### DIFF
--- a/.github/workflows/windows-arm64.yml
+++ b/.github/workflows/windows-arm64.yml
@@ -1,0 +1,37 @@
+name: Windows ARM64 installer
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - ".github/workflows/windows-arm64.yml"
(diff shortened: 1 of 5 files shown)


ogulcancelik:
@coderabbitai review

ogulcancelik:
Addressed the deterministic retry coverage: the Windows package test now holds the staged ConPTY file across activation, releases it after 300 ms, and asserts that activation succeeds through the retry. The same test retains the persistent-lock rollback case, and the latest Windows package CI passes.

Issue #3393, opened 2026-08-30
update --handoff installs the new binary before the 64-pane check, leaving the CLI protocol-locked out of the running server


phil-gusev:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

`herdr update --handoff` installs the new binary *before* the live-handoff
pane-count check runs. When the check then rejects the session, the new CLI
can no longer talk to the old server, so there is no supported way to close
panes and retry.

The binary installs, then handoff fails:
 
```
live handoff failed for session default: server live handoff failed:
{"code":"handoff_failed","message":"live handoff supports at most 64 panes in
one update; close panes or restart herdr normally"}
```
 
Rollback works correctly - the old server keeps all panes. But the installed
CLI is now protocol 20 against a protocol 19 server:
 
```
$ herdr pane list
{"id":"cli:pane:list","error":{"code":"protocol_mismatch","message":"client
protocol 20 is newer than server protocol 19; restart the Herdr server before
using this command..."}}
```

Can install old version locally to connect and close some panes and then rerun update.

### Expected behavior

 
Any one of these would resolve it:
 
1. Run the pane-count check before installing the new binary, so a rejected
   handoff leaves the install untouched.
2. Allow `pane list` / `pane close` through the protocol gate as recovery
   commands, alongside `server stop` and `status`.
3. Raise or chunk the limit.

### Reproduction

1. Run a session with more than 64 panes.
2. Run `herdr update --handoff`.

Constants responsible for it:
`MAX_FDS_PER_HANDOFF = 64` in `src/server/handoff.rs`, enforced in
`src/server/headless.rs`

### Impact

It was severely inconvenient and broke the persistent update

### Environment

- Herdr version: 0.8.0 -> 0.8.2
- Update channel (stable or preview): Stable
- Operating system: Ubuntu
- Terminal: Ghostty
- Shell, if relevant: Bash
- Relevant config, if any: None

PR #3411, opened 2026-08-30
fix(server): transfer handoff descriptors in batches

kataokatsuki:
## Problem

`herdr server live-handoff` refuses a session with 65 or more panes:

    live handoff supports at most 64 panes in one update; close panes or
    restart herdr normally

The session then stays on the old build while the CLI moves ahead, and the
only supported way out is a normal restart, which ends the shell and agent
processes in every pane.

#3393 reports the same limit from `herdr update --handoff`, where the rejected
handoff additionally leaves the freshly installed CLI protocol-locked out of
the server that is still running. This change covers only the third option
listed there, chunking the transfer; the install ordering and the protocol
check on recovery commands are untouched, so it does not close that issue on
its own.

## Cause

`send_fds` packs every pane's pty master descriptor into a single
`SCM_RIGHTS` control message and calls `sendmsg` once. `recv_fds` mirrors
that with one `recvmsg`. A single control message cannot carry an unbounded
number of descriptors, so `MAX_FDS_PER_HANDOFF = 64` guards the send, once in
`send_fds_and_wait_restored` and once in `perform_live_handoff` before any
work starts.

## Fix

- send the descriptors in batches of `FDS_PER_MESSAGE` (64)
- drop both pane-count guards
- the receive side asks for `min(remaining, 64)` descriptors per `recvmsg`
  and accumulates across calls until `expected` have arrived, iterates every
  `cmsghdr` with `CMSG_NXTHDR` and bounds each `SCM_RIGHTS` payload by
  `msg_controllen` before reading it, rejects a batch that carries more
  descriptors than it asked for, treats `MSG_CTRUNC` as unrecoverable because
  the kernel has already closed the descriptors that did not fit, and closes
  the descriptors it already holds on any failure

A session of 64 panes or fewer still produces exactly one batch, so its wire
representation is unchanged and old and new implementations stay compatible
across the range that works today. A new sender handing more than 64 panes to
an old receiver (a downgrade) still fails, but that pane count is already
unsupported. `HANDOFF_VERSION` stays at 1; raising it would refuse even the
64-and-under handoffs that currently succeed.

The sending side of a handoff is always the previously running binary, so the
first update onto a build that carries this change is still refused by the old
server's guard; batching takes effect from the next handoff after that.

64 is kept as the batch width rather than raised: a single `SCM_RIGHTS`
message accepts at most 253 descriptors on Linux and 254 on macOS, measured
with a socketpair program, and 64 leaves room under both.

I'm happy to bump `HANDOFF_VERSION` or widen the batch if you'd rather.

## Verification

Measured on base e7d82207.

- macOS, end-to-end against a real server with an isolated config, runtime
  and socket path:

  | build | panes | `server live-handoff` | panes after |
  |---|---|---|---|
  | `origin/master` | 65 | fails: `at most 64 panes` | — |
  | `origin/master` | 71 | fails: `at most 64 panes` | — |
  | this branch | 65 | completes | 65 |
  | this branch | 66 | completes | 66 |
  | this branch | 71 | completes | 71 |
  | this branch | 131 | completes | 131 |

  131 panes splits into three batches (64 + 64 + 3), so this covers more than
  the two-batch case. The script creates one pane per tab over the socket API
  and then runs `herdr server live-handoff --import-exe` against the same
  binary; "panes after" is `herdr pane list` on the replacement server.
- Linux aarch64: `cargo nextest run --locked -E 'binary(live_handoff)'`; the
  new test passes, and 5/5 when repeated alongside
  `live_server_holds_one_pty_master_fd_per_pane`.

New test `live_handoff_carries_more_panes_than_one_scm_rights_message` builds
a 70-pane session, hands off, and asserts the replacement server holds 70 pty
master descriptors and lists 70 panes.

refs #3393

### DIFF
--- a/src/server/handoff.rs
+++ b/src/server/handoff.rs
@@ -22,8 +22,11 @@ const HANDOFF_VERSION: u32 = 1;
 const READY_TIMEOUT: Duration = Duration::from_secs(30);
 #[cfg(unix)]
 const OWNED_ACK_TIMEOUT: Duration = Duration::from_millis(500);
+// Descriptors are transferred in batches of this size. A single SCM_RIGHTS
+// control message caps out at 253 descriptors on Linux and 254 on macOS, so the
+// batch stays well below both limits and the number of panes stays unbounded.
 #[cfg(unix)]
-pub(crate) const MAX_FDS_PER_HANDOFF: usize = 64;
+const FDS_PER_MESSAGE: usize = 64;
 #[cfg(unix)]
 pub(crate) const MAX_REPLAY_BYTES_PER_PANE: usize = 8 * 1024;
 #[cfg(unix)]
@@ -174,12 +177,6 @@ pub(crate) fn accept_and_validate_on(
 
 #[cfg(unix)]
 pub(crate) fn send_fds_and_wait_restored(stream: &mut UnixStream, fds: &[RawFd]) -> io::Result<()> {
-    if fds.len() > MAX_FDS_PER_HANDOFF {
-        return Err(io::Error::new(
-            io::ErrorKind::InvalidInput,
-            format!("handoff supports at most {MAX_FDS_PER_HANDOFF} pane file descriptors at once"),
-        ));
-    }
     send_fds(stream, fds)?;
 
     stream.set_read_timeout(Some(READY_TIMEOUT))?;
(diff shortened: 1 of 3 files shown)


JJLiebig (approved):

JJLiebig (commented):
Reviewed the descriptor-batching change. The ≤64-pane path stays a single `SCM_RIGHTS` message, so the wire is unchanged and `HANDOFF_VERSION` can correctly stay at 1 — bumping it would have refused handoffs that work today. The receiver accumulates across `recvmsg` calls, walks each `cmsghdr`, bounds every payload by `msg_controllen`, treats `MSG_CTRUNC` as unrecoverable, rejects over-long batches, and closes already-received fds on all failure paths; fd ordering still matches the manifest, so dropping both pane-count guards is safe. New 70-pane test passes on Linux and macOS CI. I understand this is scoped to the transfer chunking only and doesn't close #3393, since the sending binary still guards until it's on a batching build. LGTM.

kataokatsuki:
Thanks for the review and merge!

Issue #3225, opened 2026-08-25
Pi agent registration fully lost in long-running pane (agent_not_found)


formyideal:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.

### Current behavior

A long-running Pi pane **silently loses its agent registration entirely**:

- `herdr agent list` no longer includes the pane at all (another agent pane — opencode — is still listed correctly)
- `herdr list` shows the pane with no agent tag (earlier the same day it correctly showed `(current, pi, working)`)
- `herdr agent explain w6:p1 --json` returns:

```json
{"error":{"code":"agent_not_found","message":"agent target w6:p1 not found"}}
```

- The Pi TUI itself keeps running normally the whole time — only Herdr's view of it is affected
- `herdr integration status` reports `pi: current (v8)` (`C:\Users\<user>\.pi\agent\extensions\herdr-agent-state.ts`) before and after the loss

### Expected behavior

The pane should keep reporting `idle` / `working` / `blocked` for as long as the Pi session lives, or automatically re-bind after whatever event dropped the registration.

### Reproduction

No minimal repro yet — this happened organically in a long-lived session (many hours, hundreds of tool calls):

1. Start Pi inside a Herdr pane (Windows 11, PowerShell host)
2. Use the session heavily across many hours (see Environment for the extension-heavy setup)
3. At some point the pane vanishes from `herdr agent list`; `herdr agent explain <pane>` returns `agent_not_found`

Suspected contributing factors (unconfirmed, offered as leads):

- The session runs `@cortexkit/pi-magic-context`, which periodically compacts/materializes session history — possibly emitting lifecycle events similar to the `/reload` re-anchor case in #1441
- Several `pi install` operations added new packages under `~/.pi/agent/npm` during the same day (extensions load on restart, so probably unrelated, but noting for completeness)
- A daemon-side state reset combined with the extension not re-binding afterwards would match the "reports silently dropped" mechanism described in #1377 — but unlike #1231/#1377 (status frozen at `idle`), here the registration is **gone completely**

### Impact

`herdr agent list`, `herdr wait_agent`, and any event-driven supervision silently lose track of long-running Pi panes — precisely the multi-agent supervision use case Herdr is designed for.

### Workaround

Quit and relaunch Pi in the pane. The quit sends `release_agent`, and the fresh start re-binds cleanly (verified pattern from #1377).

### Environment

[...]

PR #3574, opened 2026-09-03
fix: keep an agent name through a process-exit observation

caner-akca:
## Summary

- free a pane's agent name when the agent is confirmed gone rather than on a
  single process-exit observation
- keep the name through a false exit that the next probe clears, matching how
  detection uncertainty is already handled
- leave the `pane.agent_detected { released: true }` event unchanged

This fixes the failure reported in the #3225 comment thread: a named pane stops
resolving by its name while its process keeps running, with a rename as the only
recovery. It does not address the failure in the issue body, so #3225 should stay
open.

## Validation

- `just check`
- API-boundary reproduction of the reported `agent_not_found`, plus state-layer
  tests pinning both directions — a false exit keeps the name, a confirmed exit
  frees it
- two `live_handoff` tests fail identically on `master` and this branch on macOS
  26.2, before reaching changed code: their fake agent `exec`s `/bin/sleep`, and
  `KERN_PROCARGS2` there returns an empty environment section, so `HERDR_AGENT`
  is invisible to `process_agent_hint`. Those run in CI.

refs #3225


JJLiebig (approved):
Rebased onto current master (ff32bd45) with force-with-lease; no conflicts and the diff is unchanged — 3 files, 166 insertions / 3 deletions.
Re-checked that the change still does what it should after the rebase:
- The only production change is in src/terminal/state.rs: the name is now released only when the agent is actually gone (agent.is_none() && self.recent_agent_process_exit.is_some()), instead of on a lone process-exit observation. The src/app/api*.rs edits are test-only.
- That matches the existing detection-uncertainty contract, and the added tests pin both directions — a false exit keeps the name, a confirmed exit frees it for reuse.
- Scope is explicit: this fixes the false-exit "named pane stops resolving by name" failure from the #3225 comment thread and does not address the issue body failure, so #3225 stays open.
Validation: just check green locally (2989 Rust tests + maintenance), and all 6 CI checks pass on head 27738112.

Edit: The follow up from #3225 seems to be unreachable so this should have been it, no follow up required.

PR #4158, opened 2026-09-15
fix(detect): track Muse Code 1.2.1 working state

ohk:
Muse panes missed active turns: the 0.2.1-era manifest never matched 1.2.1 chrome, so live working screens fell through to `default_known_agent_idle_fallback` (and flapped working→idle→working as the activity line danced around the 8-line window).

Personally reproduced on Herdr 0.9.0 / macOS with Muse Code 1.2.1: 6 agent panes, active turns shown as idle.

## Changes
- `working_esc_interrupt`: window 8→16 lines (the running line sits above the todo block + chrome); added approval-pair not-gates
- New `osc_title_working` (priority 800): braille spinner OSC title marks active turns even when no working line is visible; sits below all blocked rules since question UI keeps the spinner
- `idle_prompt`: added `❯` composer variants (prompt glyph changed from `⟩`); `idle_status_fallback`: added `max` effort
- Manifest version → 2026.09.15.1 (bundled + published copies in sync); CHANGELOG entry
- New `muse_manifest_detects_code_1_2_1_live_chrome` unit test with 5 live captures

## Validation
- Live working snapshot repro: idle fallback → `working / working_esc_interrupt`
- 9/9 fixtures correct on the real engine via `herdr agent explain --file`
- Live server (`reload-agent-manifests`, no restart): active panes report working with visible evidence; idle panes match `idle_prompt` with `visible_idle`; 210s of post-fix sampling with zero flaps
- nextest bin suite 3278 passed · clippy ` -D warnings` clean · fmt clean · python/bun contract gates green

---
Stack 1/5. Rest of stack: #4163 (2/5) → #4164 (3/5) → #4165 (4/5) → #4166 (5/5), merge in order. Per-layer diffs: ohk/herdr#1–#5.

PR #4164, opened 2026-09-15
feat(integration): add Muse hook install plumbing

ohk:
Stack 3/5 — merge after #4163. Per-layer diff: ohk/herdr#3 (this diff is cumulative over the stack; cross-fork PRs cannot chain bases).

First code layer of the Muse integration: session-identity hook plus install/uninstall plumbing. Session-only by design: Muse emits no Interrupt hook event and synthesizes no Stop on cancel, so lifecycle reporting would stick on working after every Esc; state stays with screen manifest plus OSC detection.

## Validation
- nextest bin suite green · fmt clean · python contracts green (clippy dead_code on unwired fns cleared by 4/5)


Answeror:
Relayed from the GChat cross-check of the 5-PR Muse stack, posted here per @ohk's request so it's actionable outside chat (stack summary on #4158).

**Bug: `Interrupt` is not a Muse hook event.** `HookEventKind` has 17 variants (`crates/types/src/hook_event_kind.rs` in Muse source) and `Interrupt` isn't one, so the `("Interrupt", None, "idle")` entry in `MUSE_HOOK_EVENTS` (`src/integration/mod.rs`) never fires. Worse, Muse never synthesizes `Stop` on cancel/interrupt, so after Esc Herdr will stick on "working" until the next prompt/tool event. Suggest dropping the dead entry and covering interrupt → idle via detection, same approach as #4163.

**Optional: `SubagentStart`/`SubagentStop` exist but aren't subscribed.** Subagent/workflow working-state currently rides entirely on #4163's terminal detection. Fine if detection holds, but the precise hook signals are available if it proves flaky.

**Correctly omitted:** `PreLLMCall`, `PostToolUseFailure`, `Notification`, `StopFailure`, etc. — none add state info beyond the `Pre`/`PostToolUse` → working bracket and `Stop` → idle.

Issue #4153, opened 2026-09-15
`herdr pane move --focus` changes server focus but does not move the attached client's viewport (0.9.0, still on master)


cjonesde:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

Running `herdr pane move <pane_id> --new-tab --workspace <target> --focus` (or `--new-workspace --focus`) from a shell inside a pane returns `outcome: ok`, and the server log shows `workspace focused` / `tab focused` for the destination workspace. The attached shell client does not follow. It stays on the origin workspace, parked on whatever tab is now that workspace's active tab once the emptied source tab closes. A few seconds later, with no CLI call in between, the server log emits another `workspace focused` / `tab focused` pair, this time back to the origin workspace, and `herdr workspace list` again reports the origin as focused.

Before 0.9.0 the same command switched the attached client to the moved pane.

### Expected behavior

The attached shell client's viewport follows the move to the destination workspace and tab, and server focus stays there, matching `herdr agent focus` after #3764 and the rule confirmed in #4014 that a successful public-socket focus request moves every attached shell client.

### Reproduction

1. Attach one shell client to a herdr session with two workspaces, `w1` and `w2`.
2. From a shell inside a pane on `w1`, run:
   ```
   herdr pane move "$HERDR_PANE_ID" --new-tab --workspace w2 --focus
   ```
3. Check the server log. It shows `workspace focused workspace_id="w2"` and `tab focused workspace_id="w2" tab_id="w2:t1"`.
4. Look at the attached client. It is still on `w1`, now showing that workspace's remaining tab.
5. Press any key in the client, then check the log again. It shows `workspace focused workspace_id="w1"` with no preceding `api request received` line.

Also reproduces through a shell wrapper that calls the same command on every `cd`: a per-project launcher runs `herdr pane move "$HERDR_PANE_ID" --new-tab --workspace <ws-for-cwd> --focus` (or `--new-workspace --focus` for a workspace that does not exist yet) whenever the target directory maps to a different workspace than the current one. Every invocation shows the same split: server log reports the destination focused, the attached client stays behind, then reverts on the next keypress.

### Impact

[...]

PR #4168, opened 2026-09-15
test: wait for runtime readiness in integration tests

minatoaquaMK2:
## Summary

Make two integration tests wait for the runtime and rendered state they actually need. This is test-only work found while validating #4153; the pane-focus fix is kept separate.

## Root cause

- The handoff test sends raw `agent.start` immediately after workspace creation. The PTY exists, but the shell may still be starting, so the request fails with `agent_pane_busy`. The same failure was reproduced with the unchanged master executable.
- The reconnect test clicks a fixed sidebar row, although that row changes depending on whether Local saved a workspace before shutdown. Its raw ANSI substring assertions also miss text delivered through incremental cell updates.

## Changes

- Queue a shell builtin that writes a readiness marker, wait for its execution, then send the original single `agent.start` request. A deliberately delayed shell exercises the startup gap.
- Locate the recovered workspace in the current rendered sidebar and assert selection and input against the reconstructed screen.
- Cover restored and unrestored workspace counts, a notice replacing the sidebar divider, and an identical label in pane output.

No production code, protocol changes, longer existing timeouts, or agent-start retries.

## Validation

- Affected handoff group, reconnect test, and helper: **25 passed**.
- Ten stress rounds of the two runtime tests and helper: **30 executions passed without retries**.
- Standalone native `just ci` with `NEXTEST_TEST_THREADS=1`: **3582 Rust tests passed, six default skips**, plus formatting, all-target Clippy, 106 maintenance, six architecture, and 39 integration-asset tests. No exclusions or retries were added.
- Windows cross-lint was not run because the Windows SDK is not installed.


JJLiebig (approved):
lgtm, thanks for the test!

PR #4171, opened 2026-09-15
fix: reconcile client locations after no-focus pane moves

minatoaquaMK2:
refs #4153

## Dependencies and scope

This supplements #4159; it is not an alternative implementation of its focused-move fix. 

#4159 and #4168 are merged, so this branch was rebased onto current `master` and now contains only the supplemental commit. The diff below is just that change.

The rebased supplemental commit, [0d7314ee](https://github.com/herdrdev/herdr/commit/0d7314ee14826b73d5d000004acfc1e91c13fa55), adds only **one production line and ten lines in an existing test**, across two files. It introduces no new test module, geometry policy, protocol change, or replacement for #4159's focus handling.

## Remaining problem

A `focus=false` move can delete the source tab while the server's numeric workspace/tab indices stay unchanged. A client viewing that tab then retains the deleted tab ID because the topology-reconciliation path does not include `PaneMove`.

Its target lookup falls back to the server default. A later, unrelated change to that default makes the client follow the wrong workspace instead of retaining the remaining source tab.

## Fix

- Classify `PaneMove` as an operation that requires client-location reconciliation.
- Extend #4159's existing focus-guard test with a second client on the source tab. After the no-focus move, change the server default and assert that this client still targets the remaining source tab.
- Retain the existing assertions for the other client's independent view, errors, unchanged moves, and zoom guards.

## Validation

- On #4159 head `1a36e5eb`: minimized regression failed **five out of five runs**, with client target `w2:t1` instead of `w1:t2`.
- With this one-line fix: **five out of five runs passed**, without retries.
- `just ci 'test(pane_move)'`: **19 Rust tests passed**, plus native lint, 106 maintenance, six architecture, and 39 integration-asset tests.
- Full native `just ci` on the final branch, including both prerequisites: **3585 Rust tests passed, six default skips**, plus formatting, all-target Clippy, 106 maintenance, six architecture, and 39 integration-asset tests. `NEXTEST_TEST_THREADS=1`; no additional exclusions or retries.
- Local Windows cross-lint was not run because the SDK is not installed.

No documentation change is needed: this preserves the existing no-focus behavior when a source tab disappears.


JJLiebig:
@coderabbitai review
@greptileai

(re-triggered after rebase onto current master to clear the prerequisite-commit conflict)

JJLiebig (approved):
The root cause and fix are correct: PaneMove was missing from shell_locations_may_need_reconcile, so a no-focus move that deletes the source tab left affected clients holding a stale tab ID that could later follow the server default; adding it to the existing reconciliation path is the minimal, targeted fix. The new second-client assertion pins the regression precisely.

LGTM

JJLiebig:
@minatoaquaMK2 thanks for all the patches Eric!

PR #4204, opened 2026-09-15
feat: simplify SSH machine setup discovery

JJLiebig:
> [!NOTE]
> **Draft vibe check:** this PR is for feedback on the product shape and terminal interaction. It is not presented as merge-ready yet.

## Why

`herdr machine add` currently makes users supply a display label and know the remote session name. On Windows, SSH `PATH` can also resolve an older binary while a compatible server is already running from another executable, leading setup toward an unnecessary install prompt and exposing expected probe noise.

## Proposed flow

```text
herdr machine add <ssh-target>
              |
              +-- --remote-session supplied --> use it explicitly
              |
              `-- discover running sessions
                    |-- none --> use default
                    |-- one  --> select automatically
                    `-- many --> interactive picker
                                 (non-interactive: require --remote-session)
                              |
                              v
             choose display label
                    |-- --label supplied --> use it, no prompt
                    |-- interactive ------> edit a prefilled default
                    `-- non-interactive --> accept the default
                              |
                              v
             validate known remote executables
             PATH -> running Herdr process -> managed install
                              |
                              v
             prepare server -> save machine
```

Multiple-session interaction:

```text
$ herdr machine add workbox
Running sessions on workbox:
> default
  agents

↑/↓ select · Enter confirm · Esc cancel
Machine label: workbox
```

The selected session uses both `>` and bold text. The label is a real editable prefill with the cursor at the end, so typing `-test` produces `workbox-test`; Enter accepts it and Esc or Ctrl+C cancels. Named sessions default to `workbox/session-name`.

A code block is intentional for now: this is a text-only CLI flow, so it shows the interaction more clearly than a screenshot.

## First-pass behavior

- Makes `--label` optional. Interactive callers can edit a prefilled label; `--label` skips the prompt, and non-interactive callers use the computed default.
- Discovers running sessions through the existing `session list --json` command.
- Selects one running session automatically and opens a bold Up/Down picker when several are running.
- Preserves `--remote-session` for explicit and non-interactive callers.
- On Windows, adds executables from running Herdr processes to the existing candidate list and validates capabilities before reuse; it does not scan the filesystem.
- Filters expected POSIX-probe and successful PowerShell stderr so `/bin/sh` and CLIXML noise do not leak into successful setup. SSH authentication notices and failed-command diagnostics remain visible.

## Deliberately deferred

- Adding several sessions from one machine in a single operation.
- Per-session compatibility annotations and version presentation.
- Rich onboarding or SSH-management UI.
- Any broader executable or filesystem discovery.

## Validation

- Real `jjdesktop` flow: validated automatic selection and a two-session picker with Down/Esc, raw-mode restoration, no saved profile on cancellation, compatible executable reuse, and no `/bin/sh` or CLIXML noise.
- Real PTY label edit: `jjdesktop` became `jjdesktop-test`; Ctrl+C cancelled cleanly and the catalog remained empty.
- Focused picker, label-editor, parsing, discovery, and probe tests passed.
- `just lint` passed.
- Full nextest suite excluding the unrelated, independently reproducible Windows ACL fixture failure: 3,114 passed, 7 skipped.
- 112 maintenance tests passed (5 skipped); UI architecture and integration asset tests passed.
- `just check` is blocked only by `platform::windows::config_backup::tests::backup_preserves_legacy`.
- Current-head GitHub CI passed on Ubuntu, macOS, and Windows, including Windows ConPTY packaging and the Unix SSH-authentication visibility contract.

Discussion: https://github.com/herdrdev/herdr/discussions/4198


JJLiebig:
@greptileai
@coderabbitai
let's do a lil review

Issue #3769, opened 2026-09-08
Remote attach UI becomes sluggish in 0.9.0


cirog:
## What happened

Since updating to Herdr 0.9.0, remote Herdr sessions are much less responsive for me. When I use Herdr remotely, clicks can take a few seconds to register. Refreshing and general interaction also feel slow.

This was not the case before 0.9.0. The same remote workflow was responsive before the update.

The problem is specific to remote attach. If I run `herdr --remote <host>` from my local machine, the session is slow and clicks sometimes appear to do nothing. If I SSH to the same machine and run `herdr` there, it is fast.

## Environment

* Herdr client: 0.9.0
* Herdr remote server: 0.9.0
* Channel: stable
* Platform: Linux ARM64 client and server
* Connection: SSH via `herdr --remote`

## What I checked

The local remote bridge is running and uses almost no CPU. The remote server also responds quickly to CLI and API requests. This looks specific to the remote client, bridge, or rendering path.

The local client log contains repeated warnings like this:

```text
flushing lone escape after input timeout; if this follows an alt chord or focus switch it may reach the pane as plain esc bytes
```

I do not see obvious remote disconnect or timeout errors.

## Expected behavior

Remote Herdr should remain responsive. This worked before 0.9.0.

## Actual behavior

Remote clicks and refreshes can be delayed by several seconds, making the session feel temporarily frozen.

cirog:
I found some relevant logs from both sides.

On Helios, the client log shows successful 0.9.0 endpoint handshakes for the remote sessions:

```text
2026-09-08T05:29:31.342660Z endpoint handshake succeeded generation=1 server_version=0.9.0
2026-09-08T06:04:26.222047Z endpoint handshake succeeded generation=1 server_version=0.9.0
2026-09-08T07:00:53.519612Z endpoint handshake succeeded generation=1 server_version=0.9.0
2026-09-08T08:45:34.860349Z endpoint handshake succeeded generation=1 server_version=0.9.0
2026-09-08T09:20:02.708187Z endpoint handshake succeeded generation=1 server_version=0.9.0
```

The same log also records two transport losses earlier in the morning:

```text
2026-09-08T08:08:37.084030Z endpoint transport failed endpoint=local error=connection was lost
2026-09-08T09:08:07.612682Z endpoint transport failed endpoint=local error=connection was lost
```

These say `endpoint=local`, so I am not sure they are directly related to the remote slowdown.

[...]

Issue #3745, opened 2026-09-08
0.9.0 regression: terminal popup typing over SSH is slower and sends ~234 KB per key


huilang021x:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

After upgrading both ends from 0.8.2 to 0.9.0, typing in a terminal popup over SSH feels delayed; ordinary panes remain responsive.

Reproduced with a plain Bash input loop, without the original worktree plugin. On the same MacBook → Linux SSH connection (30 keystrokes per case, 299×80 terminal, no artificial rate limit):

| Version / input | Echo p50 | Echo p95 | Downstream bytes/key, median |
|---|---:|---:|---:|
| 0.8.2 pane | 8.36 ms | 15.79 ms | 89 |
| 0.8.2 popup | 9.05 ms | 12.54 ms | 92 |
| 0.9.0 pane | 5.42 ms | 6.81 ms | 94 |
| 0.9.0 popup | 35.35 ms | 73.14 ms | 234,270 |

Latency is from writing a key to the Mac PTY until Herdr emits its echo, excluding physical display/IME latency. Byte counts are Herdr payload before SSH encryption, not SSH packet sizes.

### Expected behavior

Popup typing should remain responsive, as it was on 0.8.2, without this large increase in per-keystroke traffic.

### Reproduction

1. On Linux, create executable `/tmp/herdr-popup-bench/echo.sh`:

```bash
#!/bin/bash
printf 'BENCH_READY\n'
while IFS= read -e -r -p 'BENCH_INPUT> ' line; do
  printf 'ACCEPTED\n'
done
```

2. In an isolated server config, set `terminal.default_shell` to that script and add:

```toml
[keys]
prefix = "ctrl+b"
[[keys.command]]
key = "prefix+shift+g"
type = "popup"
command = "/tmp/herdr-popup-bench/echo.sh"
width = "82%"
height = "72%"
```

3. Connect from the Mac using matching client/server versions and server keybindings. Type in the ordinary pane, then press prefix+G and type in the popup. Compare 0.8.2 with 0.9.0 at the same window size.

[...]

PR #4209, opened 2026-09-15
fix: keep remote surface updates incremental

ogulcancelik:
## problem

small remote screen changes can still resend the whole surface when popups, graphics, or metadata prevent the existing cell-only patch. on a slow link this makes typing and streaming output lag.

## change

- negotiate optional `endpoint.surface-delta.v1` without changing frozen generation-1 codecs or dropping older peers
- send changed main/popup cell spans with current metadata; reuse already delivered kitty assets
- choose a delta only when its framed bytes are smaller than the full surface and within existing message limits
- preserve wire baselines for ordinary recomputation; keep hard resets for lost presentation state
- bound collection allocations during decoding and commit baselines/assets only after successful enqueue

## validation

- rebased onto `fc548768`; `just check`: 3,624 rust tests passed, plus maintenance, architecture, integration assets, docs contracts, and windows msvc lint
- `just bench-render-scale`: all six profiles passed, including sparse/dense/checkerboard cases with 1/15 panes and 1/15 clients
- regression coverage includes popup and graphics deltas, metadata, negotiation/fallback, frozen encoding, malformed/count-amplifying input, and transactional baselines
- ssh rohan comparison against the original master base: at 250 kb/s, popup 120x40 went from 35,553 to 494 bytes/update and p95 306 to 97 ms; popup 195x50 went from 67,858 to 494 bytes/update and p95 439 to 102 ms
- static kitty image + text: 58,089 to 524 bytes/update; the candidate sent the asset initially without resending it for each text change
- rohan topology smoke passed: 15 tabs, active split, throttled second client, navigation, resize, selection, and reconnect
- captured 15 independent opencode conversations using a free model and replayed identical terminal bytes/timing into both builds: master 3.05% vs patch 3.00% combined herdr server/client cpu, effectively unchanged; all final pane text matched

## limits and tradeoff

synthetic checkerboard damage still costs more cpu to save relatively few bytes. the post-rebase profile measured roughly 26–31% more encode/decode cpu than reuse mode for 5–7% fewer bytes; dense updates also have candidate-selection overhead. this is not claimed to be a universal cpu improvement.

opencode used four 20-second matched replay windows, one client, and 15 split panes; 13 conversations streamed answer text while two mostly waited/reasoned. rohan used 12 samples per case. those live/replay results were collected before the final rebase; full checks and scaling benchmarks were rerun afterward. this is not exhaustive workload or image-visual certification.

refs #3769
refs #3745


ogulcancelik:
@coderabbitai review
@greptileai

rebased onto fc548768; the surface implementation is unchanged. full checks and render scaling passed again on deada5e9.

Issue #4109, opened 2026-09-14
Session navigator selects the workspace instead of the matching pane after search


cead22:
## Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

## Current behavior

After searching in the session navigator, Herdr highlights the workspace context row instead of the matching pane row.

For example, searching for `exclusive` shows a workspace named `PR reviews` and a matching pane named `Exclusive lock`. Herdr highlights `PR reviews`. Pressing Enter focuses the workspace. I must press Down once before pressing Enter to focus the matching pane.

The matching pane is visible, but Herdr does not select it automatically.

## Expected behavior

Herdr should select the first matching target after a search.

When a pane matches the query, the workspace should remain visible as context, but the pane row should be highlighted. Pressing Enter should focus that pane directly.

When only the workspace matches, Herdr should continue selecting the workspace row.

## Reproduction

1. Open Herdr `0.9.0` with a workspace containing a pane titled or labeled `Exclusive lock`.
2. Press `Ctrl+B`, then `G` to open the session navigator.
3. Press `/` to focus search.
4. Type `exclusive`.
5. Observe that the workspace row is highlighted while the matching `Exclusive lock` pane row is not.
6. Press Enter. Herdr focuses the workspace.
7. Press Down, then Enter. Herdr focuses the matching pane.

## Impact

Search does not provide the expected direct navigation. Every pane search requires an extra Down keypress, and pressing Enter first opens the broader workspace instead of the requested pane.

## Environment

- Herdr version: 0.9.0
- Update channel (stable or preview): stable
- Operating system: macOS arm64
- Terminal: iTerm.app
- Shell, if relevant: zsh
- Relevant config, if any: none

The same search flow selected the matching pane on Herdr `0.8.2`.

PR #4223, opened 2026-09-16
fix: select matching targets in session navigator search

minatoaquaMK2:
refs #4109

## Problem

Editing a session-navigator query clears its explicit selection. The client then defaults to the first visible row, even when that row is a workspace or machine retained only as context for a matching pane. Pressing Enter focuses the context instead of the search target.

## Changes

- Keep the existing text-match result on client-only navigator rows and select the first matching row for a nonempty query.
- Keep matching workspaces, tabs, and machines eligible in tree order rather than always preferring panes.
- Preserve explicit mouse/keyboard selection of context rows. Home explicitly selects the first visible row, while search-editor Home/End retain cursor-editing behavior.
- Leave empty queries, state-only filters, row visibility, and endpoint-qualified focus behavior unchanged.

No server state, persistence, API, or wire-format changes.

## Validation

- The new regression failed before the fix: searching `exclusive` selected `Workspace(ws_1)` instead of `Pane(pane_2)`. It now checks both the rendered highlight and the Enter-generated pane-focus request.
- All 15 navigator tests pass, covering workspace/tab/pane and cwd matches, case/whitespace normalization, pasted and edited queries, manual context selection, Home, and remote endpoint-qualified targets.
- `just check`: 3631 tests passed, 6 skipped; native lint, maintenance, integration assets, documentation contracts, and Windows cross-target lint passed.
- Additional Windows `cargo clippy --all-targets --locked --target x86_64-pc-windows-msvc -- -D warnings` passed, including test-code compilation.
- `just bench-render-scale` passed before and after the fix, including background-workspace and active-pane profiles. Navigator measurements at fixed 106x30 geometry:

| Populated panes | Before (us/frame) | After (us/frame) |
| --- | ---: | ---: |
| 1 | 371.3 | 365.3 |
| 15 | 396.7 | 395.9 |
| 52 | 408.4 | 402.8 |

The 1-to-15-pane increment was 25.4 us before and 30.6 us after; total frame times remained essentially unchanged. Matching predicates are reused rather than recomputed.


minatoaquaMK2:
@coderabbitai review
@greptileai

minatoaquaMK2:
@coderabbitai review
@greptileai

minatoaquaMK2:
@coderabbitai full review

minatoaquaMK2:
Closing this as superseded by #4384.

I checked current master (`d59d0603`) using the original regression from this PR, without applying this PR’s implementation. Searching `exclusive` now highlights the matching pane, and Enter sends the correct pane-focus request. All 17 navigator tests passed.

The new Go To design addresses the original reproduction in #4109, so this PR no longer needs to be rebased or merged. Thanks for the navigation improvements!

Issue #4178, opened 2026-09-15
`events.subscribe` silently drops events when more than ~500 events occur faster than they are delivered; no gap indication


kennyzheng-builds:
### Is this a reproducible bug?
- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior
A subscriber that is connected and reading continuously does not receive all matching events after a burst. Some events are missing from the middle of the stream. No error frame is sent, the connection stays open, and the envelopes carry no sequence number, so the subscriber cannot tell that anything was lost.

Running the reproduction below exactly as written (`nc -U` subscriber reading continuously, then a plain `for` loop of 600 `herdr workspace rename` calls, which took about 3 s, all returning `ok`) on an otherwise idle server:

- received **465 of 600** `workspace_renamed` events, in order;
- labels ran `flood-0` .. `flood-526`, with **62 holes inside that range**, and `flood-527` .. `flood-599` were never delivered;
- no error frame; the connection stayed open.

Events arrive at a steady ~10 per second per subscription, so a backlog builds up after any burst.

### Expected behavior
Either every matching event emitted after `subscription_started` is delivered in order, or the subscriber is told that events were dropped (for example an explicit "events lost" frame, or a sequence number on each envelope that makes gaps detectable), so it can resync from `session.snapshot` / `agent get`.

### Reproduction
```sh
# 1. subscriber (keep open, print every line)
{ printf '%s\n' '{"id":"s1","method":"events.subscribe","params":{"subscriptions":[{"type":"workspace.renamed"}]}}'; sleep 120; } \
  | nc -U "$HERDR_SOCKET_PATH" | tee events.txt

# 2. in another terminal: a burst of 600 renames
herdr workspace create --label flood          # note the workspace id, e.g. w3
for i in $(seq 0 599); do herdr workspace rename w3 "flood-$i" >/dev/null; done

# 3. after ~60 s
grep -o 'flood-[0-9]*' events.txt | wc -l      # fewer than 600 (465 in our run)
grep -o 'flood-[0-9]*' events.txt | head       # numbering has holes
```

### Impact
[...]

PR #4224, opened 2026-09-16
test: wait for app readiness before late session injection

minatoaquaMK2:
## Problem

`session_appearing_after_startup_is_preserved_before_autosave` waits for the API socket and then injects its supposedly late session file. The listener is bound before `App` evaluates restore, so the test can instead put the file in place during normal startup and fail because no recovery backup was needed.

This blocked full CI while validating #4178 and reproduced with the unchanged master executable.

## Change

Add a read-only `workspace list` round trip through `App` and assert that it returns an empty workspace list before injecting the file.

This is **three added lines in one existing test**. The fixture, autosave deadline, byte-for-byte backup assertion, production code, and other tests are unchanged. There are no sleeps or retries.

## Validation

- Original test fails identically with unchanged master and the separate #4178 build.
- With the readiness barrier: **five independent runs passed**, without retries.
- Full standalone native `just ci`: **3624 Rust tests passed, six default skips**, plus formatting, all-target Clippy, 112 maintenance, six architecture, and 39 integration-asset tests. No retries or additional exclusions.
- Local Windows cross-lint was not run because the SDK is unavailable.


JJLiebig (approved):
Nice one, thanks Eric.

Issue #3819, opened 2026-09-09
macOS: case-only session name can delete a different named session


gauravyad86:
## Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

## Current behavior

On a case-insensitive filesystem, Herdr accepts a named session recorded as `Foo`. After it is stopped, `herdr session delete foo --json` succeeds and removes the persisted `Foo` session directory. The response identifies the deleted session as `foo`, even though `herdr session list --json` listed only `Foo`.

## Expected behavior

A delete command with a different-case name must not remove another named session. It should preserve `Foo` unless the target matches the recorded session name exactly.

## Reproduction

In a temporary config directory on default macOS APFS:

```sh
tmp=$(mktemp -d)
XDG_CONFIG_HOME="$tmp/config" herdr --session Foo server
XDG_CONFIG_HOME="$tmp/config" herdr session list --json
# lists a stopped named session named Foo
XDG_CONFIG_HOME="$tmp/config" herdr session delete foo --json
# returns {"deleted":true,..."name":"foo"...}
test ! -e "$tmp/config/herdr-dev/sessions/Foo" && echo removed
```

The last command prints `removed`.

## Impact

A case-only typo in `session delete` can erase saved state for a different named session, including its workspace and pane restore data.

## Environment

- Herdr version: 0.9.0 checkout build (`b9ce96869e89`)
- Update channel (stable or preview): N/A (checkout build)
- Operating system: macOS 26.5.2, arm64; default case-insensitive APFS
- Terminal: local terminal shell
- Shell, if relevant: zsh
- Relevant config, if any: fresh temporary `XDG_CONFIG_HOME`

hdosys:
I implemented a focused fix for this in our v0.9.0-based fork: [implementation and tests](https://github.com/hdosys/herdr-win/commit/150503d0192475563d94ddda79fa92c087c573bf).

`delete_session` now resolves the requested name against the immediate session-directory entries using exact `OsStr` equality. It uses only the matched entry's path for the running-session check and deletion. If there is no exact match, it returns an actionable error without attempting removal. Enumeration errors also propagate.

This avoids globally lowercasing session names, while preserving the reserved-default and running-session protections.

[...]

PR #4233, opened 2026-09-16
fix: require exact session names before deletion

minatoaquaMK2:
On a case-insensitive filesystem, `herdr session delete foo` currently deletes the saved session listed as `Foo`. Reproduced with Herdr 0.9.0 on macOS 27 (Apple Silicon) in a temporary configuration directory; the new regression test also fails against unmodified master because the saved data disappears.

Resolve the exact directory-entry spelling before probing the session socket or deleting data. Reject case aliases with guidance to use `herdr session list`, while preserving deletion of the exact stopped session, missing-session idempotency, and the default/running-session protections. Add regression coverage and update the unreleased English, Chinese, and Japanese docs.

Validation:
- Native `just ci` stages passed as part of `just check`: formatting, Clippy, 3448 tests, maintenance/architecture checks, and integration asset tests.
- `just docs-contract-test` passed separately.
- Patched CLI verified: `delete foo` preserves `Foo`, `delete Foo` succeeds, and repeated deletion succeeds.
- Full `just check` stopped at Windows cross-lint because the local Windows SDK configuration is unavailable; no Windows validation is claimed. The case-sensitive sibling test is included for filesystems that support distinct `Foo` and `foo` directories.

Refs #3819.


minatoaquaMK2:
@coderabbitai review
@greptileai

Please review the latest commit, aecf5f509d61cf19b0cd023b19ebacb592adb62b. Both previous findings are addressed: the session deletion regressions now run through isolated CLI subprocesses without changing the parent process environment, and a Drop guard cleans up each temporary directory even after an assertion panic. A dedicated panic-cleanup regression and the ordinary parallel test harness both pass; just ci passed with 3449 tests.

JJLiebig (approved):
Thanks Eric, lgtm!

Issue #3763, opened 2026-09-08
machine add fails after SSH MFA on 0.9.0


zxdvd:
### Is this a reproducible bug?

- [x] I confirm this is a reproducible bug, not a feature request, idea, question, contribution proposal, or direction check.
- [x] I reproduced this bug on the version and environment reported below using the exact steps provided.

### Current behavior

`herdr --remote jms-dev` works with my JumpServer MFA setup. But `machine add` asks for an OTP, then fails after I enter a valid code:

```text
error: lost connection to server: server closed connection; machine was not saved
```

This happened on repeated attempts. The remote Herdr server was already running and reported compatible.

### Expected behavior

Adding the same host should work after entering a valid MFA code.

### Reproduction

1. Use a JumpServer SSH alias requiring a public key followed by keyboard-interactive TOTP. Leave Herdr's SSH settings at their defaults, with no shared `ControlPath` in SSH config.
2. Run `herdr --remote jms-dev` and verify that it connects, then detach.
3. Run `herdr machine add jms-dev --label jms-dev` and enter the current OTP. It fails with the error above.

### Impact

I can use remote attach, but can't add this host to the multi-machine view.

### Environment

- Herdr version: 0.9.0 locally and remotely (unmodified release)
- Update channel (stable or preview): stable
- Operating system: macOS 15.5 arm64; remote Linux
- Terminal: Ghostty
- Shell, if relevant: zsh locally
- Relevant config, if any: `PreferredAuthentications publickey,keyboard-interactive`, `KbdInteractiveAuthentication yes`, `IdentitiesOnly yes`

aureliolapa:
I was having the same issue, than I read the docs: 

> If authentication fails, check ordinary SSH first. For a passphrase-protected key, load it with ssh-add before starting Herdr’s non-interactive background connections.

Then I run  ssh-copy-id <target> so the destination has the key of the source. And now is working

johnuopini:
I have the same issue when using 1password SSH agent, standard remote works fine, machine add doesn't, the local agent has no keys, its a native OSX keystore integration, key is loaded dynamically, normally this doesn't cause any issue

johnuopini:
My 1Password comment above was wrong, ignore it.
Here the bridge connection fails on host key checking. herdr passes StrictHostKeyChecking=yes on the ssh command line, and that wins over the config file, where I have UserKnownHostsFile /dev/null so no host key is ever stored.
[...]

PR #4276, opened 2026-09-17
fix(remote): reuse ssh authentication for saved machines

kazunari-kamata:
## Summary

Make saved SSH machine setup work after password or keyboard-interactive authentication, and allow `herdr --remote` launched inside a Herdr pane to start an independent remote client.

## Root Cause

`machine add` establishes an interactive SSH control connection during setup, but its endpoint-capability probe started a separate `BatchMode=yes` SSH bridge without that managed control socket. The probe could not answer the password or MFA prompt and returned a generic connection-closed error, so the machine profile was not saved.

Separately, SSH commands launched from a Herdr pane inherit Herdr runtime environment variables. SSH configurations that forward those variables make the remote bridge appear nested and cause it to abort.

## Changes

- Reuse the managed interactive SSH control socket for the endpoint probe that follows setup.
- Remove inherited Herdr runtime and socket override environment variables from every SSH path that launches the remote bridge.
- Add focused regression tests for control-socket reuse and runtime-environment isolation, including the bridge path.

## Docs

No public docs update is needed. This corrects SSH setup and runtime behavior without changing documented commands, configuration, or workflows.

## Validation

- `just ci`
  - 3,471 tests passed; 6 skipped.
  - 122 maintenance tests passed.
  - UI hot-path and integration-asset checks passed.
- Manual verification: macOS arm64 to macOS x86_64 saved-machine setup using password authentication completed successfully.
- Manual verification: macOS arm64 to Windows x86_64 saved-machine connection completed successfully after configuring SSH key authentication.


kazunari-kamata:
Addressed the bridge-path environment inheritance finding in a69f4bec. The shared runtime-environment scrub now applies to both `RemoteSsh::base_command` and `bridge_ssh_command`, with dedicated bridge-path coverage. `just ci` passed (3,471 tests; 6 skipped, plus maintenance and integration checks).

JJLiebig:
@coderabbitai
@greptileai 
full review pls

JJLiebig (changes requested):
The control-socket reuse is the right small fix for the setup probe, and the shared SSH environment cleanup is reasonable. I traced setup, the local client launch, saved reconnects, session selection, and teardown on 9f0cdae0. Please address the following before approval:

1. **Complete the local remote-client launch isolation.** `run_client_process` (src/remote/attach.rs:3170) still launches `herdr client` with inherited `HERDR_ENV=1`. `main` handles that child through `exit_if_nested_disabled` before connecting. Consequently, `herdr --remote` from a Herdr pane still fails with the default configuration even after both SSH commands are scrubbed. I reproduced the child rejection on this head with an isolated configuration and socket path. Clear the inherited marker for this specific client launch while preserving the bridge socket, and add coverage for that launch path. Please also correct the root-cause explanation: on the current code, `remote-client-bridge` dispatches before the nested guard; the demonstrated rejection is in the local `client` child. Keep ordinary local nested-session protection intact.

2. **Preserve the requested remote session when forwarding environment variables.** The new scrub omits `HERDR_SESSION`. `machine add` is dispatched before `configure_from_args`, so a named local pane can supply that variable even when the saved profile requests the default remote session. With matching SSH `SendEnv HERDR_*` / server `AcceptEnv` settings, `RemoteExecutable::session_args` omits `--session default`, and remote status/startup/bridge commands then select the inherited name. I verified that `HERDR_SESSION=local-named-session` selects that session, whereas an explicit `--session default` selects the default socket. Complete the environment isolation or make the remote selection explicit, and cover default versus named remote sessions under an inherited local session. This is a remaining gap in the isolation behavior advertised by the PR, not a claim that the PR introduced the original inheritance bug.

3. **Make the authentication regression exercise the production connection path.** `endpoint_probe_reuses_interactive_ssh_control_socket` only checks that `endpoint_probe_ssh_options` returns `ssh.options()`. Changing the real `SshStdioBridge::start` call back to `None` would leave it green. Replace that helper-only assertion with coverage observing the managed configuration/control socket reaching the actual probe SSH command, including its noninteractive setting. The regression should fail when the production call loses those options. The one-line forwarding helper can then be removed. Likewise, the environment tests need to protect the local child and session cases above, not just the two SSH command constructors.

One scope limitation to make explicit in the description/validation: the authenticated setup connection is temporary. `RemoteSsh::drop` sends `ssh -O exit`; saved connections later use `new_noninteractive` without those managed options. This fixes saving the profile after authentication, but does not by itself enable subsequent password/MFA-only connections without a separately usable noninteractive authentication path. Please distinguish successful setup from successful saved-client connection/reconnection. I am not asking this PR to introduce persistent credentials or redesign connection ownership. Windows clients also retain their documented lack of managed Unix control-socket reuse.

Validation: all 88 local `remote::` tests passed on Windows, plus the isolated nested-client and session-selection probes above. I have not run a live password/MFA SSH setup. The changes otherwise remain focused; I found no need for a broader transport or protocol redesign.

kazunari-kamata:
Thank you for the review. I am working through the feedback and will report back with the completed changes and validation results.

kazunari-kamata:
Thank you again for the detailed review. Addressed in fa8c5bab:

- The local `herdr client` launched by `herdr --remote` now removes `HERDR_ENV` while retaining its explicitly supplied bridge socket.
- SSH bridge commands also remove `HERDR_SESSION`, preventing a forwarded local session from affecting the remote process.
- The endpoint-probe regression test now verifies the bridge command carries both the managed control socket and `BatchMode=yes`.
- The PR description now clarifies that password/MFA authentication is only used by the interactive setup connection and is not stored; later reconnects require independent non-interactive SSH authentication.

I also fixed the two pre-existing Rust 1.97 Clippy diagnostics so the project lint check passes.

kazunari-kamata:
Addressed the macOS CI failure in 162283e7 by synchronizing the slow-observer test before the writer starts. This removes the startup scheduling race while preserving the assertion that ongoing writes reset the timeout. The focused test passed five consecutive local runs.

kazunari-kamata:
Closing this PR because the upstream implementation in #4490 now addresses #3763 with managed SSH authentication reuse and recovery commands. It overlaps with this PR\u2019s endpoint-probe change and conflicts during rebase.

This PR also contains an additional nested-runtime environment isolation change for `herdr --remote`, but that behavior has not been independently adopted upstream. Rather than carry an overlapping patch on top of the official implementation, I am closing this PR.

PR #4338, opened 2026-09-18
test: focus agent detection coverage on engine contracts

ogulcancelik:
## Summary

Remove 45 tests that classify frozen CLI screens or titles against individual agents’ bundled detection rules. Those tests cannot establish compatibility with a changing CLI.

Keep engine and integration contracts: manifest validation, matching and priority, loading and reloads, process identification, and hook/protocol behavior. Add three synthetic engine tests for independent OSC inputs and production dispatch, skip-state semantics, and structural region extraction. Decouple the fallback assertion from bundled rules.

Update AGENTS.md to require live, version-recorded smoke tests for agent-specific detection changes instead of screen/title fixture tests.

No runtime code or detection manifests changed.

## Validation

- Focused detection suite: 96 tests passed.
- `just check`: 3,623 Rust tests passed, plus maintenance, integration, documentation checks, and Windows lint.


ogulcancelik:
i’m leaving out the suggested pi-specific dispatch test. identification mappings remain tested, and the new synthetic-manifest test exercises the production dispatch path without bundled screen assumptions. the registry-completeness gap you noted is distinct and could use a general structural test, but i’m not adding another per-agent dispatch expectation in this cleanup.

PR #4340, opened 2026-09-18
perf: enable ssh compression for remote connections

ogulcancelik:
## Summary

Remote scrolling can keep replaying old screen updates after the user reverses direction or stops. Request SSH compression for Herdr connections to reduce that backlog on bandwidth-limited links.

- Add `-C` through the shared SSH and SCP option helpers, including unmanaged connections and the initial connection that establishes a multiplexed master.
- Keep the existing Herdr protocol, codecs, and server compatibility unchanged.
- Document that `-C` overrides `Compression no`, while an existing external ControlMaster keeps its original transport settings.

## Reproduction and measurements

Used installed Herdr 0.9.1 client/server processes in a disposable named session, a rootless loopback OpenSSH server, and a userspace ProxyCommand throttling **encrypted TCP bytes**. The same 160×50 alternate-screen fixture receives 240 wheel events at 120/sec, reversing halfway, then a typed sentence with cursor edits. App-side timestamps distinguish input arrival from client-side visible updates.

These are actual OpenSSH compression results, not the earlier compression simulation:

| Connection | Compression | Scroll reversal | Catch-up after scrolling stops |
|---|---|---:|---:|
| 4 Mbps, 80 ms RTT | off | 5,037 ms | 5,177 ms |
| 4 Mbps, 80 ms RTT | on | 119 ms | 116 ms |
| 1 Mbps, 80 ms RTT | on | 166 ms | 127 ms |
| unrestricted | off | 33 ms | 30 ms |
| unrestricted | on | 30 ms | 28 ms |

All wheel events arrived and the final edited text matched exactly. Input arrived promptly even when pictures were delayed. These are single runs of a synthetic, redraw-heavy text workload, not a claim about every app, graphics, or network. Runtime tests exercised the same SSH compression setting with installed binaries; command-construction tests cover the candidate's flag wiring.

## Validation

- Rebased onto current `origin/master` before opening.
- `just check`: 3,623 Rust tests passed, plus maintenance, integration, docs, Linux lint, and Windows-target lint.
- Independent reviewer: no blockers; documentation clarification applied.
- Independent simplifier: no changes recommended.

No dependencies or server/client wire changes. Compression reduces traffic; it does not provide a general bound on outstanding screen updates.


JJLiebig:
@coderabbitai review
@greptileai

mlamp:
Heads-up for users behind **Tailscale SSH**: the `-C` from this PR is silently a no-op there. Tailscale's built-in SSH server (`remote software version Tailscale`) negotiates no compression:

```
$ ssh -C -v <tailnet-host> true 2>&1 | grep compression
debug1: kex: server->client cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
debug1: kex: client->server cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
```

Any host with `tailscale up --ssh` intercepts tailnet :22, so `herdr --remote <tailnet-host>` stays uncompressed even with this change. Reaching the host's OpenSSH on another port (or via its non-tailnet address) is the workaround.

Data point from a high-RTT link where it matters (0.9.1 client and server, Bangkok → Helsinki, ~216 ms RTT, direct WireGuard path, no DERP):

- Switching to a busy workspace sends ~140 KB of surface data; it arrives in ~12–20 KB steps per RTT, ~2.3 s until the pane repaints. Scrolling is worse, since each step waits on the previous `pane.scroll` response.
- Same session via `ssh -t host herdr` (server-side UI) feels roughly 10× faster.

It might be worth a line in the remote-attach docs, or a warning when the negotiated compression is `none`.

JJLiebig:
@mlamp that is a great flag, would you mind adding an issue for that? seems like something we should add

PR #4362, opened 2026-09-18
fix(client): make the event loop un-parkable and pace the headless render loop

vietairs:
## Why

On macOS 27.2 beta the herdr **client** hangs the host terminal (cmux + Terminal.app) with a
"Not Responding" dialog. Root cause (proven from hang report 17314: 2.19 h unresponsive, all 8
threads parked): the client event loop parked in its `tokio::select!` with a **far timer
deadline + silent channels**, and `should_quit` was polled only at the loop head — so **Ctrl-C
could neither wake nor exit it**. A secondary pressure source was the fixed 16 ms render cadence.

## What changed (7 files, +182/−12)

**1. Quit-wake arm — the loop can no longer park forever**
- New `ClientLoopEvent::Quit`; the `ctrlc` termination handler (SIGINT/SIGTERM/SIGHUP) now
  `should_quit.store(true, Release)` **and** `quit_notify.notify_one()`.
- `quit_notify.notified()` is a first arm of the select on both the unix (biased) and windows
  paths. A `tokio::Notify` permit is retained across iterations, so a termination signal is
  never lost even if it lands between loop passes. The `Quit` arm re-stores the flag and lets the
  existing shutdown path run — no new teardown path.

**2. Timer clamp — bounded re-wake**
- `ClientLoopTimer::deadline()` clamps every requested delay to `MAX_LOOP_TIMER_DELAY` (250 ms)
  **before** the sticky-min. Any far/stale deadline re-wakes the loop (quit-flag check + health
  tick) at least every 250 ms. Normal delays are unaffected; an earlier sticky deadline still wins.

**3. Local-endpoint heartbeat**
- Local **shell** connections now participate in the existing ping/pong health probe: a
  silent-but-alive local server surfaces as a 10 s timeout → clean `ConnectionLost` instead of an
  undetectable wedge.
- Local **terminal-attach** connections (no shell surface) stay exempt — they cannot false-expire.
  The server already echoes pongs for every probed connection (verified end-to-end).

**4. `ui.render_interval_ms`**
- New config key (default 16, clamped to ≥ 1) that is the single source of the
  render/presentation pacing interval previously hard-coded as `MIN_RENDER_INTERVAL`. Bounds the
  headless render loop's cadence without a rebuild; reloadable live via `[ui]`.

## Scope notes
- **Deferred (out of the minimal ANR closure, logged):** I2 client-side timeout on terminal-open
  await; I6 kitty-graphics flush pacing; I7 server-side dead-client reaper (the server still can't
  detect a *silent* client); I8 ops tooling.
- The `render_interval_ms` knob paces the **headless (server) render loop** specifically; the
  client-side TUI shell render path is paced by the 250 ms timer clamp instead.
- SIGHUP now runs the full quit path (intended — a terminal hangup should exit a TTY-attached
  client; documented in-code).

## Testing
- `cargo check` clean (0 new warnings).
- Focused tests **130/130 pass** across the four changed subsystems (`client::timer`,
  `client::endpoint`, `config::model`, `app::runtime`); new tests pin the clamp, the local
  probe/exempt behavior, and the render-interval parse/clamp/reload.
- Untouched-module failures are **pre-existing flakiness** (clean base fails 7; with this change 5;
  strict subset — zero new failures).

## Merge note (AD-04)
Base branch `macos-anr-mitigation` is cut from `v0.9.0-hvn.2` (`792b8cf4`), which is now ==
`origin/master`, so this is a clean single-commit diff into `master`. No auto-merge.


vietairs:
Accidental push: this branch belongs to the fork vietairs/herdr, not upstream. Closing; the change is being PR'd in the fork.

PR #4389, opened 2026-09-19
fix: stabilize windows cursor redraws and settle deadlines

JJLiebig:
Windows ConPTY can finish a synchronized redraw at a temporary cursor position, then restore the input caret in a later write. This causes the drawn cursor to flash above the Codex composer.

This is a separately maintained follow-up to #4310, preserving Can's original commits and adding the native-Windows-tested follow-up. #4310 and its branch are left open and unchanged for maintainer choice.

- Ignore intermediate synchronized-frame positions and preserve eligible cursor candidates across idle gaps.
- Hold row changes and moves larger than two columns for up to 100 ms; ordinary caret steps retain the 20 ms window.
- Remove the follow-up's redundant burst timer, which could postpone cursor movement indefinitely when writes arrived 30 ms apart.
- Schedule the final repaint after the applicable hold, so a cursor moves even when output stops. Preserve that deadline when graphics output shares the write.

Validation:
- Native Windows 11 build 26200.9457, Windows Terminal 1.24.260710001, Codex 0.155.1-fork.1 with gpt-6-astra low and animated composer.
- System and bundled ConPTY: typing, Home/End and left/right navigation stayed on the composer row across 73 and 74 visible cursor frames respectively. No model prompt submitted.
- Both backends passed synthetic 10/30/70 ms delayed caret restoration and final-position publication with no further output.
- Regression checks demonstrated failures before the timer/repaint corrections. Full native `just check` passed again on rebased head `0e89e28e`, including formatting, clippy, Rust tests, maintenance/integration/docs checks, and build.
- Two independent local reviewers reported no findings on the rebased head.

Limits: the 100 ms hold remains a timing heuristic and also delays legitimate large caret moves. Live observations cover PTY/server frames through a nested client, not high-speed recordings of Windows Terminal pixels or every agent/editor. Live smoke tests preceded the rebase; cursor logic was unchanged by the rebase. No native macOS qualification. Temporary sessions were removed; existing sessions preserved.

refs #4303


JJLiebig:
@axelalbertkarlsson Could you give this one a try?

I spent a fair bit more time tinkering with this and this feels quite good but want a second machine input if possible.

Not just letting it run but also how it feels overall while working with it, (if possible)!

axelalbertkarlsson:
@JJLiebig I gave this a try with both `codex-cli 0.156.0-alpha.7` and `codex-cli 0.155.1`, and this feels really good.

I’m not seeing any jumps away from the input line, either while Codex is working or while it’s idle.

I also made some attempts based on [#4310](https://github.com/herdrdev/herdr/pull/4310), trying to address the behavior from [#4303](https://github.com/herdrdev/herdr/issues/4303). I could stop the caret from jumping, but it made the UI feel sluggish.

The latest changes in here don’t seem to have that tradeoff: no notable jumps, still feeling responsive.

I’ve only tested it with relatively small things so far, I'll test it further, but from what I’ve seen this looks like it resolves the jumping issue without introducing the sluggishness. Good job!

JJLiebig:
@axelalbertkarlsson thank you for that!

The approach here is to not guard cursor redraw across the board but only when it jumps further out. When I was testing a more generic approach it just didn't feel good at all so figured this was a good idea to test.

If there's any blockers you notice let me know, I'll play around with it for another hour or two and then call it good enough until the reworked animations land for codex in 0.156.*

axelalbertkarlsson:
@JJLiebig Smart! I’ve been using it for the last hour or so and haven’t noticed anything weird.

If you need a second pair of eyes to test any other issues, just ping me, happy to help.

